PR #27, opened 2026-07-31
Pull local models 3

### LINKED ISSUE #3: feat: Pull model for local providers
## Description

In order to actually use a local model, it first needs to be pulled which is a different operation for each provider. We need to add a new trait method to `pull_model` with a given `ModelMetadata` that will do the backend-specific logic to pull the model in a way that the provider can use. We then need to implement it for each of the existing providers with default no-op behavior for hosted providers. The generic OpenAI provider should warn on pull that it can't pull and may need to be pulled manually using the logic for the specific OpenAI API provider.

gabe-l-hart:
## Description

Closes #3 

This PR adds `granite-cli model pull` support for all current backends.

### DIFF
--- a/src/commands/model.rs
+++ b/src/commands/model.rs
@@ -4,9 +4,10 @@ use anyhow::Result;
 // Local
 use crate::commands::ProviderCommands;
 use crate::dependency::{self, Configured, DependsOn, Requirement};
-use crate::models::{ContextFit, MODEL_REGISTRY, ModelMetadata, ModelType};
+use crate::models::{ContextFit, MODEL_REGISTRY, ModelMetadata, ModelType, ModelVariant};
 use crate::utils::hardware::detect_hardware;
-use crate::providers::{Provider, ProviderMetadata, ProviderSource};
+use crate::providers::{Provider, ProviderMetadata, ProviderSource, ProviderType, PROVIDER_REGISTRY};
+use crate::utils::ui::Ui;
 use crate::utils::Searchable;
 
 /// Compare semantic versions in descending order (higher versions first).
@@ -440,7 +441,7 @@ impl ModelCommands {
 
                 let model_config = crate::config::ModelConfig {
                     model_id: model_id.to_string(),
-                    provider_id,
+                    provider_id: provider_id.clone(),
                     variant: Some(format!("{}/{}", selected_variant.format, selected_variant.precision)),
                     enabled: true,
                 };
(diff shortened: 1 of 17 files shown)


src/providers/base.rs
Tanker2020, line 1: The default `pull_model()` implementation returns `Ok(())`, which makes the contract of this trait method pretty ambiguous. From the caller’s perspective, `Ok(())` reads like “the model was successfully pulled”, but for providers that don’t actually implement pull/download behavior this can mean “nothing happened”.

That shows up in `model pull`, where the command delegates to the provider and can end up reporting success even if the provider cannot actually perform a pull automatically.

I think this should be modeled more explicitly so success always means the same thing. A few possible ways to tighten this:
- return a dedicated `UnsupportedPull` / `ManualActionRequired` error for providers that can’t do automatic pulls
- add a capability check like `supports_pull()`
- or make `pull_model()` required for implementers instead of giving it a silent-success default

Right now the abstraction makes it hard for callers to distinguish “download completed” from “provider doesn’t support this”.

src/providers/openai.rs
Tanker2020, line 1: This method returns `Ok(())` even though it only warns that the user must pull the model manually.

That creates a misleading success path for `model pull`: the CLI command can appear to succeed even though no model weights were downloaded.

I think this should return an error (or some explicit unsupported/manual-action result) instead of success. Otherwise scripts and users have no reliable way to tell whether the pull actually happened.

src/providers/ollama.rs
Tanker2020, line 1: I think this has a false-success case.

The implementation only treats a streamed `"success"` status as the terminal success signal, but if the NDJSON stream ends early for any reason (server restart, connection interruption, truncated response, etc.), control falls through to `pull_finish(..., None)` and returns `Ok(())`.

That would mark an interrupted/incomplete pull as successful.

Could we track whether a terminal success event was actually observed, and if the stream ends before that, return an error instead? Something like “pull stream ended before completion” would make this much safer.

src/commands/model.rs
Tanker2020, line 1: At the CLI level, I think it would help to separate these cases more explicitly:

1. provider pulled successfully  
2. provider supports pull but pull failed  
3. provider does not support automatic pull and requires manual setup

Right now those states can collapse into the same success-looking path depending on the provider implementation. Tightening that distinction would make the command much safer for both users and automation.

gabe-l-hart: Good call. I'll rework this into a more expressive state description.

gabe-l-hart: done

src/providers/ollama.rs
gabe-l-hart: Good catch! Fixed

src/providers/openai.rs
gabe-l-hart: Good call. Fixed with the refactor on valid responses.

PR #30, opened 2026-08-03
chore: No warnings

### LINKED ISSUE #29: chore: No warnings
## Description

There are currently a number of warnings during build. Many are caused by unused code paths that are anticipated as the `launch` verb gets fleshed out, but we should handle those correctly with marks so that they don't pollute the build and hide legitimate warnings that we want to pay attention to.

The goal of this issue is to have a clean build with no visible warnings.

gabe-l-hart:
## Description

Closes #29 

This PR clears out all warnings in both build and test. The warnings were split between legitimate issues and warnings for unused pieces that will eventually be needed once the architecture is fully fleshed out (particularly from `define_factory!`).


src/capabilities/base.rs
```diff
@@ -148,177 +146,3 @@ pub struct ToolConfig {
 use crate::define_factory;
 
 define_factory!( Capability, CapabilityMetadata, CapabilityFactory);
-
-/*-- Capability Matching Logic ------------------------------------------------*/
```
gabe-l-hart, line 152: This stuff was all dead and will need to be rewritten down the road anyway, so just removing it completely for now.

src/commands/capability.rs
```diff
@@ -95,7 +95,7 @@ impl CapabilityCommands {
 
                 // Check dependencies
                 ctx.ui.info("Checking dependencies:");
-                let mut all_satisfied = true;
+                // let mut all_satisfied = true;
```
gabe-l-hart, line 98: These code paths were dead pending implementation of correct dependency behavior, so fully commenting out for now.

src/commands/model.rs
```diff
@@ -838,7 +838,7 @@ mod tests {
     }
 
 
-    fn metadata_supporting(formats: Vec<ModelFormat>, precisions: Vec<&str>) -> ProviderMetadata {
```
gabe-l-hart, line 841: This arg never got removed when `precisions` stopped being part of `ModelMetadata`

Issue #19, opened 2026-07-31
feat: Capability architecture


gabe-l-hart:
## Description

This issue is to define and implement the core `Capability` architecture. This includes finalizing `trait Capability` and `CapabilityMetadata` to support the proposed initial Capability set. It also includes defining the expected `Dependency` flow so that individual capabilities can declare dependencies on specific models and/or providers. It may also include requiring dependencies on specific installed commands on the host system (eg `uvx`).

Tanker2020:
## CapabilityType enum dependency from PR #33

While implementing the Launcher architecture (PR #33), @gabe-l-hart flagged that the `supported_capabilities` field on both `LauncherMetadata` and the `Launcher` trait currently uses `Vec<String>`. Once capabilities are fleshed out, this should be migrated to:

```rust
// Proposed change in launchers/base.rs
fn supported_capabilities() -> HashSet<CapabilityType>;
```

Key points:
- Individual launchers should declare their dependency on specific **types** of capabilities (e.g. `MCP`, `Skill`, `SubAgent`, `Model`), not capability instance IDs
- Using `HashSet` over `Vec` guarantees uniqueness and avoids ambiguity of duplicate values
- This provides abstract separation between individual capability implementations and launchers, so capabilities and launchers are loosely coupled
- The tricky case noted was Mellea Skills Compiler, which runs at config time and needs agent-specific knowledge (e.g. "find all current skills") — this may need a separate lifecycle hook beyond just declaring a type dependency

This change is blocked on the `CapabilityType` enum being defined here in the capability architecture. PR #33 will be updated once that enum exists.

PR #33, opened 2026-08-04
Feat/launcher architecture

### LINKED ISSUE #20: feat: Launcher architecture
## Description

This issue is to define and implement the core `Launcher` architecture. This includes defining `trait Launcher` and `LauncherMetadata` to support the proposed initial `Launcher` set. It also includes defining the expected configuration flow so that individual launchers can declare support for specific `Capabilities`. It will also require that launchers are able to validate the presence of specific commands on the host machine (allowing users to provide manual path overrides for non-PATH installation locations).

Additionally, users should be able to create multiple configurations for the same launcher with name overrides. This would allow, for example `claude-local` (run against a local model), `claude-enterprise` (run against an enterprise model router), and `claude` (run against factory claude) to co-exhist. The default should be to create a name that matches the launcher type name (eg `"claude"` for `CaludeLauncher`, `"bob"` for `BobLauncher`). When running setup for a given launcher, if there is an existing launcher that matches the type name, the user should be given the option to update the existing config or create a new named launcher using the given type.

Tanker2020:
# feat: Launcher architecture

Closes #20

## What this PR does

Implements the core **Launcher** abstraction — a first-class system for registering, configuring, and executing external AI coding tools (Claude CLI, Bob CLI, and others to come) with a Granite env-var overlay injected into the subprocess.

A Launcher is the *outward-facing* complement to a Provider. A Provider answers "which inference server does granite-cli talk to?" A Launcher answers "which external tool do I exec, and how do I configure it?" The two sit on opposite sides of the system: providers are reached over HTTP by granite-cli itself; launchers are wrapped processes that then talk to their own provider endpoint via the env overlay granite-cli injects.

---

## Background — what are launchers?

Launchers are AI coding agent CLIs that a developer already has installed on their machine — tools like [Anthropic's Claude Code](https://docs.anthropic.com/en/docs/claude-code) (`claude`) or IBM Bob (`bob`). They are not part of granite-cli; granite-cli wraps them.

Intended usage:

```bash
granite-cli launch claude -- --resume
```

Granite-cli resolves the binary, builds an environment overlay (e.g. `ANTHROPIC_BASE_URL=http://localhost:11434` pointing at a local Granite model), spawns the tool as a subprocess with those env vars set, then waits and propagates the exit code. The tool itself is completely unmodified — it simply sees different environment variables.

---

## Files changed

| File | Action | Purpose |
|---|---|---|
| `src/launchers/base.rs` | new | `trait Launcher`, `LauncherMetadata`, `LaunchContext`, `EnvBinding`, `LauncherFactory` via `define_factory!` |
| `src/launchers/claude.rs` | new | `ClaudeLauncher` + `ClaudeLauncherConfig` — wraps the `claude` binary |
| `src/launchers/bob.rs` | new | `BobLauncher` + `BobLauncherConfig` — wraps the `bob` binary |
| `src/launchers/mod.rs` | new | `LAUNCHER_REGISTRY` static, `LauncherSource` implementing `Configured<dyn Launcher>` |
| `src/commands/launcher.rs` | new | `LauncherCommands` — `catalog`, `list`, `setup`, `validate`, `remove` |
| `src/config/mod.rs` | extended | `LauncherConfig` struct, `launchers` field on `Config`, `launchers/` persistence dir, full CRUD helpers |
| `src/commands/mod.rs` | extended | Re-exports `LauncherCommands` |
| `src/main.rs` | extended | Adds `Commands::Launcher` subcommand tree; replaces the `Commands::Launch` "Phase 3" stub with a working `run_launch()` |
| `src/utils/ui/app.rs` | extended | Adds `Section::Launchers` to the TUI — catalog browse, detail pane, nav count, tab-cycle |
| `Cargo.toml` | extended | Adds `which = "7"` for cross-platform PATH resolution |

---

## Architecture — following existing patterns

The Launcher layer is a direct structural parallel to the existing Provider layer. Every design decision mirrors what Providers already do:

| Concern | Provider (existing) | Launcher (this PR) |
|---|---|---|
| Core trait | `trait Provider` | `trait Launcher` |
| Metadata | `ProviderMetadata` | `LauncherMetadata` |
| Factory | `define_factory!(Provider, …)` | `define_factory!(Launcher, …)` |
| Registry static | `PROVIDER_REGISTRY` | `LAUNCHER_REGISTRY` |
| Config struct | `ProviderConfig` | `LauncherConfig` |
| Config persistence | `~/.config/granite-cli/providers/` | `~/.config/granite-cli/launchers/` |
| Dependency resolution | `ProviderSource : Configured<dyn Provider>` | `LauncherSource : Configured<dyn Launcher>` |
| CLI commands | `commands/provider.rs` | `commands/launcher.rs` |

---

## Intentional divergences from the Provider pattern

Two places where Launchers behave differently from Providers are called out explicitly in code comments so future contributors understand the intent.

### 1. Binary validation is not called at construction time

Provider construction is always cheap — it just sets up an HTTP client. `Launcher::validate_command()` does a filesystem or PATH lookup, which would make `launcher catalog` and `launcher list` fail on machines where the tool isn't installed. So `validate_command` is a separate explicit call, not wired into `ConfigConstructable::new`.

### 2. Type-aware instance clash detection in `setup`

Provider setup only checks whether the exact instance ID already exists. Launcher setup additionally scans for any configured launcher with the same `launcher_type` — regardless of instance name — and offers to either update that existing entry or proceed with the new name. This prevents users accidentally creating two configs for the same tool (e.g. running `setup claude --id claude2` when `claude` already exists).

---

### Setup wizard — binary path fallback

When a user runs `launcher setup claude` and the `claude` binary is not on their PATH, the wizard does not fail immediately. Instead it asks:

```
  Warning: 'claude' not found on PATH.
  Enter the full path to the binary (e.g. /usr/local/bin/claude): _
```

It patches that path into the config JSON, constructs a second launcher instance with the override, and re-validates. Only if the explicit path also doesn't exist does it bail. Config is never saved with a non-functional binary.

### launch command — dry-run mode

The existing `Commands::Launch` stub ("Tool launching will be available in Phase 3") is replaced with a working implementation. The `--dry-run` flag prints exactly what would be exec'd and what env vars would be injected, without spawning anything:

```
$ granite-cli launch my-claude --dry-run -- --resume
Would exec: /usr/local/bin/claude
  args: --resume
  env overlay: (none)
```

### Exit code propagation

The spawned tool's exit code is propagated to the shell. A non-zero exit from the tool causes `granite-cli launch` to exit non-zero with a message, so scripts calling `granite-cli launch` get correct failure signalling.

### Remove

Launcher instances can be fully removed: `granite-cli launcher remove <id>` deletes the YAML file from `~/.config/granite-cli/launchers/`, removes the in-memory entry, and confirms. After removal, `launcher list` no longer shows the entry and `granite-cli launch <id>` returns an actionable error.

---

## TUI — Launchers section added, launch not added

The interactive TUI gains a **Launchers** section (Tab twice from Models) showing the registered launcher types, their default commands, and a detail pane. This follows the same read-only browse pattern as the existing Models, Providers, and Capabilities sections.

**Why the TUI does not offer launch:** The TUI runs inside ratatui's alternate screen mode — it has exclusive control of the terminal's raw input and output buffers. Spawning a subprocess that itself expects a full interactive terminal (which both `claude` and `bob` do) would require tearing down the alternate screen, handing the terminal to the child process, waiting, then re-entering alternate screen. This is non-trivial to do correctly across platforms, creates jarring visual transitions, and makes error recovery fragile.

More fundamentally, it isn't the right UX: the TUI is a *browse and inspect* surface. Launching a tool is an action, not an inspection, and belongs at the shell prompt where the user already has full terminal context.

A reasonable future improvement would be a TUI shortcut (e.g. pressing `Enter` on a configured launcher instance) that prints the correct `granite-cli launch <id>` command to the status bar or copies it to the clipboard — giving the user the command to run without granite-cli trying to manage the terminal handoff itself.

---

## Tests — 336 passing (up from 299)

| Module | Tests added | What they cover |
|---|---|---|
| `launchers::base::tests` | 6 | Factory register/construct, validate_command errors, env_overlay default, metadata Display |
| `launchers::claude::tests` | 5 | command_name default, explicit path override, validate_command error, metadata, config schema |
| `launchers::bob::tests` | 5 | Same coverage for BobLauncher |
| `launchers::tests` | 5 | Registry lookup, enabled-only construction, unknown type skipped, catalog contents, config schema |
| `config::tests` | 3 | LauncherConfig YAML round-trip, launchers dir created, insert/remove lifecycle |
| `commands::launcher::tests` | 13 | catalog columns/contents, list columns/sort/sentinel, validate errors, setup clash detection, setup unknown type, remove lifecycle |
| `utils::ui::app::tests` | 2 updated | Tab-cycle tests updated for 6 sections (was 5); launcher tab position assertion added |

---

## What is not in this PR

- **Capability hooks at launch time** — `on_pre_launch` / `on_shutdown` on `Capability` are defined but not yet wired through `env_overlay`. `LaunchContext.base_env` is the designed injection point; deferred until concrete capability implementations exist.
- **The `configure <tool>` wizard** — binding capabilities to a launcher during setup is a separate issue.
- **Provider failover at launch time** — selecting a working provider before handing its URL to the tool is deferred.
- **TUI launch action** — see note above; tracked as a future improvement.


gabe-l-hart (dismissed):
This is a test!

src/main.rs
```diff
@@ -211,6 +227,39 @@ enum ProviderSubcommands {
...
+
+        /// Nickname for this launcher instance. Defaults to `launcher_type`;
+        /// pass a distinct value to configure multiple named instances of
+        /// the same catalog type (e.g. `--id claude-local`).
+        #[arg(long = "id")]
+        instance_id: Option<String>,
+    },
+
+    /// Validate that the launcher's binary is reachable
+    Validate {
```
gabe-l-hart: I'd like to avoid holding this as a standalone subcommand if at all possible. I think we should be able to have the validation happen only with the actual `launch` action, but will know more once I'm deeper in.

src/main.rs
```diff
@@ -211,6 +227,39 @@ enum ProviderSubcommands {
...
+        /// Configured launcher instance ID to validate
+        launcher_id: String,
+    },
+
+    /// Remove a configured launcher instance
+    Remove {
```
gabe-l-hart, line 251: Probably not for this PR, but we should have this for all types.

src/config/mod.rs
```diff
@@ -117,6 +118,33 @@ pub struct ConfiguredCapability {
...
+    pub launcher_id: String,
+    #[serde(rename = "type")]
+    pub launcher_type: String,
+    /// Absolute path to the binary for non-PATH installs. `None` means use PATH.
+    pub command_path: Option<PathBuf>,
+    /// Capability IDs enabled for this launcher instance.
+    pub enabled_capabilities: Vec<String>,
+    /// Launcher-type-specific config passed to `ConfigConstructable::new`.
+    pub config: serde_json::Value,
+    pub enabled: bool,
```
gabe-l-hart: I know we have `enabled` in several other places, but I think it's pretty worthless, so we should probably remove it everywhere.

src/launchers/base.rs
```diff
@@ -0,0 +1,228 @@
...
+/// Core trait for launcher implementations.
+/// All launchers must implement this trait along with ConfigConstructable.
+#[async_trait]
+pub trait Launcher: ConfigConstructable + Send + Sync {
+    fn name(&self) -> &str;
+
+    /// The binary/command this instance will exec.
+    /// Concrete impls return their default name (e.g. `"claude"`); the command
+    /// path override in config is applied by `validate_command`.
+    fn command_name(&self) -> &str;
```
gabe-l-hart: NIT: I think we might just want to call this `command` to indicate that it's the full command (including path if needed).

src/launchers/base.rs
```diff
@@ -0,0 +1,228 @@
...
+
+    /// Capability IDs this launcher instance supports.
+    fn supported_capabilities(&self) -> Vec<String>;
```
gabe-l-hart, line 30: I've been torn about which direction this knowledge should flow. I think we'll need to tackle this when we really flesh out capabilities. My gut is that we'll need some kind of enum of capability types (MCP, Skill, SubAgent, Model) and individual launchers would declare their dependency on specific types of capabilities and know how to plug a given type into its config surface. This would provide some abstract separation between the individual capability implementation and the launcher.

The one that I have the hardest time figuring out would be something like Mellea Skills Compiler which runs at config time and needs agent-specific knowledge (eg "find all current skills").

src/launchers/base.rs
```diff
@@ -0,0 +1,228 @@
...
+    /// The default implementation covers the common case: resolve binary,
+    /// build overlay, merge with current process env, spawn, wait.
+    /// Concrete launchers override only when they need non-standard behaviour.
+    async fn launch(
+        &self,
+        args: &[String],
+        ctx: &LaunchContext,
+    ) -> anyhow::Result<std::process::ExitStatus> {
+        let binary = self.validate_command()?;
+        let overlay = self.env_overlay(ctx).await?;
```
gabe-l-hart, line 63: I think we may need to codify a standard way to have file-based config that gets overlaid as well. We can probably tackle that when we come to it (or individual launchers can override `launch` to support it), but I'm thinking we may need a place in the `granite-cli` HOME where we can write out ephemeral config files that we then point to using env vars.

src/launchers/base.rs
```diff
@@ -0,0 +1,228 @@
...
+}
+
+/// Metadata describing a launcher implementation (type-level, catalog entry).
+#[derive(Debug, Clone, Serialize, Deserialize)]
+pub struct LauncherMetadata {
+    pub name: String,
+    pub description: String,
+    /// The default binary name used for PATH lookup (e.g. `"claude"`, `"bob"`).
+    pub default_command: String,
+    /// Capability IDs this launcher type can make use of.
```
gabe-l-hart, line 130: If we go the route of `CapabilityType`, I think we'll want to change the type here to `HashSet<CapabilityType>` (better to use a set to guarantee uniqueness and avoid the ambiguity of multi-present values.

src/launchers/base.rs
```diff
@@ -0,0 +1,228 @@
...
+impl std::fmt::Display for LauncherMetadata {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        write!(f, "{}", self.description)
+    }
+}
+
+/// Runtime context passed through the launch lifecycle.
+pub struct LaunchContext {
+    pub launcher_id: String,
+    pub working_dir: PathBuf,
```
gabe-l-hart, line 146: What is this `working_dir` set to? I'm guessing I'll find it elsewhere, but a comment would be good to disambiguate between "the CWD of the command when it runs" and "the place the launcher can put temporary stuff"

src/launchers/base.rs
```diff
@@ -0,0 +1,228 @@
...
+
+/// A single environment variable binding contributed to the subprocess overlay.
+pub struct EnvBinding {
+    pub key: String,
+    pub value: String,
+}
+
+/*-- private --*/
+
+use crate::define_factory;
```
gabe-l-hart: I think we should put this at the top for consistency

src/launchers/claude.rs
```diff
@@ -0,0 +1,126 @@
...
+use std::path::PathBuf;
+
+/*-- public --*/
+
+#[derive(Debug, Clone, Serialize, Deserialize, Default, schemars::JsonSchema)]
+pub struct ClaudeLauncherConfig {
+    /// Override path to the `claude` binary for non-PATH installs.
+    /// Leave unset to use PATH lookup.
+    #[serde(default)]
+    pub command_path: Option<String>,
```
gabe-l-hart, line 15: How does this differ from the `command_path` in `LauncherConfig`?

src/config/mod.rs
```diff
@@ -117,6 +118,33 @@ pub struct ConfiguredCapability {
     pub config: HashMap<String, String>,
 }
 
+#[derive(Debug, Clone, Serialize, Deserialize)]
+pub struct LauncherConfig {
+    pub launcher_id: String,
+    #[serde(rename = "type")]
+    pub launcher_type: String,
+    /// Absolute path to the binary for non-PATH installs. `None` means use PATH.
+    pub command_path: Option<PathBuf>,
```
gabe-l-hart: I think this will get confused with the type-specific generic json config below (eg `ClaudeLauncherConfig.command_path`).

src/launchers/claude.rs
```diff
@@ -0,0 +1,126 @@
...
+    fn name(&self) -> &str {
+        "Claude CLI"
+    }
+
+    fn command_name(&self) -> &str {
+        self.config.command_path.as_deref().unwrap_or("claude")
+    }
+
+    fn supported_capabilities(&self) -> Vec<String> {
+        vec![]
```
gabe-l-hart: I think we should be able to invoke `self.metadata().supported_capabilities()` here to avoid needing to double-encode the list of supported capabilities

src/launchers/claude.rs
```diff
@@ -0,0 +1,126 @@
...
+            if p.exists() {
+                return Ok(p);
+            }
+            anyhow::bail!("explicit path '{}' does not exist", p.display());
+        }
+        which::which("claude").map_err(|_| anyhow::anyhow!("'claude' not found on PATH"))
+    }
+
+    async fn env_overlay(&self, _ctx: &LaunchContext) -> anyhow::Result<Vec<EnvBinding>> {
+        Ok(vec![])
```
gabe-l-hart, line 48: Ok, this makes sense. We'll need to flesh this out a bunch once we've got actual capabilities. Ideally it will be generic over `CapabilityType`.

src/commands/launcher.rs
```diff
@@ -0,0 +1,531 @@
...
+                ctx.ui
+                    .info(&format!("Available types: {}", available.join(", ")));
+                anyhow::bail!("Launcher type not found");
+            }
+        };
+
+        ctx.ui
+            .info(&format!("\nSetting up launcher: {launcher_type}"));
+        ctx.ui.info(&launcher_def.description);
+        ctx.ui.info(&format!(
```
gabe-l-hart, line 104: We need to make this configurable. Right now it just prints the default, but doesn't let the user update it.

src/launchers/bob.rs
gabe-l-hart, line 1: Same patterns here as with claude, so we'll need to update here as well

src/commands/launcher.rs
```diff
@@ -0,0 +1,531 @@
...
+                anyhow::anyhow!("No config schema registered for launcher type '{launcher_type}'")
+            })?;
+        let defaults = ctx
+            .config
+            .get_launcher(&instance_id)
+            .map(|lc| lc.config.clone())
+            .or_else(|| LAUNCHER_REGISTRY.default_config(launcher_type))
+            .unwrap_or_else(|| serde_json::json!({}));
+
+        let mut config = prompt_from_schema(&*ctx.ui, &schema, &defaults)?;
```
gabe-l-hart, line 175: I _think_ this should be what triggers the option to override the command, but I'm not seeing it work for the current implementation types.

src/commands/launcher.rs
```diff
@@ -0,0 +1,531 @@
...
+                ctx.ui.info(&format!("  Binary found: {}", path.display()));
+            }
+            Err(_) => {
+                // Binary not on PATH — ask for an explicit path
+                ctx.ui.warn(&format!(
+                    "  '{}' not found on PATH.",
+                    launcher_def.default_command
+                ));
+                let explicit = ctx.ui.text(
+                    "  Enter the full path to the binary (e.g. /usr/local/bin/claude): ",
```
gabe-l-hart: NIT: We might want to use something less specific than `/usr/local/bin/claude`

src/commands/launcher.rs
```diff
@@ -0,0 +1,531 @@
...
+                        );
+                    }
+                }
+            }
+        }
+
+        let launcher_config = crate::config::LauncherConfig {
+            launcher_id: instance_id.clone(),
+            launcher_type: launcher_type.to_string(),
+            command_path: None,
```
gabe-l-hart: Seems like this one is never used!

src/commands/launcher.rs
```diff
@@ -0,0 +1,531 @@
...
+        }
+
+        Ok(())
+    }
+
+    /// Validate that the configured launcher's binary is reachable.
+    ///
+    /// Prints the resolved absolute path on success, or a clear error message
+    /// on failure. Useful for diagnosing non-PATH installs.
+    pub fn validate(ctx: &crate::AppContext, launcher_id: &str) -> Result<()> {
```
gabe-l-hart: It seems like we shouldn't need this as a standalone command. We do explicit validation at `setup` time, so the only worry is that the binary stops existing (moved, deleted, etc). We can re-do validation at launch time and possibly even at `list`/`info` time, so I don't think we need a standalone command for this.

gabe-l-hart (changes requested):
Another great feature! I've got a few refactors I'd like to see, some of which we might want to punt until we flesh out `Capabilities`:

1. Keep the type-agnostic `LauncherConfig` as slim as possible (there are several unused fields now)
2. Make `supported_capabilities` operate over an enum of `CapabilityType`s so that we can loosely couple capabilities and launchers (NOTE: This will require starting the capabilities overhaul, but probably worth doing now)
3. No need for a standalone `validate` command
4. Need to figure out why command path overrides aren't working at `setup`

src/main.rs
Tanker2020: Done — `Validate` variant removed from `LauncherSubcommands` and `LauncherCommands::validate()` deleted. Validation now happens at `setup` time (already present) and at `launch` time via `validate_command()` in `run_launch`.

src/launchers/base.rs
Tanker2020: Captured in the comment on issue #19 alongside the enum discussion. No change here yet since `HashSet<CapabilityType>` depends on `CapabilityType` being defined in the capability architecture first.

src/launchers/base.rs
Tanker2020: Added doc comment: `/// The working directory for the spawned subprocess (i.e. the CWD the tool process will see). Typically set to the user's current directory.`

src/launchers/base.rs
Tanker2020: Done — renamed `command_name()` → `command()` on the `Launcher` trait and all implementations (`ClaudeLauncher`, `BobLauncher`, `FakeLauncher`). Updated doc comment to clarify it returns either a bare binary name for PATH lookup or a full absolute path when set by the user.

src/config/mod.rs
Tanker2020: Removed from `ModelConfig`, `ProviderConfig`, `CapabilityConfig`, `ConfiguredCapability`, and `LauncherConfig`. Cleaned up all call sites — `list` tables no longer have an `ENABLED` column, `LauncherSource` and `ProviderSource` no longer filter on it, and the `capability setup` wizard no longer prompts for enable/disable.

src/launchers/base.rs
Tanker2020: Agreed. Left a detailed comment on issue #19 capturing the `CapabilityType` enum requirement, the `HashSet<CapabilityType>` change, and the Mellea edge case. Will pick this up once the capability architecture lands.

src/launchers/base.rs
Tanker2020: `use crate::define_factory` is now the first statement under `/*-- private --*/`, before `define_factory!(...)`.

src/main.rs
Tanker2020: Tracked in GitHub issue #36 — "feat: Validate launcher binary at launch and list time".

src/launchers/base.rs
Tanker2020: Tracked in GitHub issue #37 — "feat: File-based config overlay for launchers". Individual launchers can override `launch` in the interim until a standard pattern is codified.

src/config/mod.rs
Tanker2020: Resolved by removing `LauncherConfig.command_path` — there is now a single source of truth for command path in the type-specific config (`ClaudeLauncherConfig`, `BobLauncherConfig`).

src/commands/launcher.rs
Tanker2020: Fixed. Root cause: `prompt_from_schema` returned `""` for an unset `Option<String>` field, which deserialised as `Some("")` — causing `validate_command` to try `PathBuf::from("")` and fail into the fallback prompt. Added a normalisation step (`"" → null`) before construction so the schema prompt is the single place to set a path. Removed the redundant second-chance `text()` prompt.

src/commands/launcher.rs
Tanker2020: Removed — `launcher validate` is gone from both the CLI and the commands module. See reply on the `src/main.rs:251` thread.

src/launchers/claude.rs
Tanker2020: Done — `supported_capabilities()` on both `ClaudeLauncher` and `BobLauncher` now delegates to `Self::metadata().supported_capabilities` instead of hard-coding `vec![]`.

src/launchers/bob.rs
Tanker2020: All fixes applied to `BobLauncher` in parallel with `ClaudeLauncher` — `command()` rename, `supported_capabilities()` delegation to metadata, test names and example paths updated.

src/commands/launcher.rs
Tanker2020: The old fallback prompt with that hardcoded example has been removed entirely. Test fixtures that referenced `/usr/local/bin/claude` and `/usr/local/bin/bob` have been updated to `/opt/bin/claude` and `/opt/bin/bob`.

src/commands/launcher.rs
Tanker2020: Fixed. The info line now reads `"Default command: claude (leave command_path blank to use PATH lookup)"` and the schema prompt immediately follows, letting the user set `command_path` to any value. An empty string is normalised to `null` before construction so leaving it blank means PATH lookup.

src/commands/launcher.rs
Tanker2020: This was the `command_path: None` field in the `LauncherConfig` struct literal. That field has been removed from `LauncherConfig` entirely — resolved as part of slimming down the struct.

src/launchers/claude.rs
Tanker2020: Agreed — no change here for now. The `env_overlay` default returns `vec![]` and will be wired up once concrete capability implementations exist and `CapabilityType` is defined.

src/launchers/claude.rs
Tanker2020: Removed the top-level `LauncherConfig.command_path` entirely — there is now only one `command_path`, living in `ClaudeLauncherConfig`. The `list` command reads it from `cfg.config["command_path"]`.

Tanker2020:
Addressed all four items from the review:

1. **`LauncherConfig` slimmed down** — removed both `enabled` and the top-level `command_path`. The struct now only holds `launcher_id`, `launcher_type`, `enabled_capabilities`, and `config`.
2. **`CapabilityType` enum** — deferred as agreed; left a detailed comment on issue #19 with the full design (enum variants, `HashSet<CapabilityType>`, Mellea edge case).
3. **Standalone `validate` command** — removed from CLI and commands module entirely.
4. **Command path override at `setup`** — fixed. Root cause was `prompt_from_schema` returning `""` for an unset `Option<String>`, which deserialised as `Some("")` and caused `validate_command` to fail silently into a redundant fallback prompt. Added a normalisation step (`"" → null`) before construction so the schema prompt is the single place to set a path.

src/main.rs
gabe-l-hart: I've also created https://github.com/ibm-granite-community/granite-cli/issues/38 which tracks adding `remove` for the other types

src/main.rs
```diff
@@ -399,3 +449,87 @@ async fn run_configure(ui: &dyn Ui, _args: ConfigureArgs) -> anyhow::Result<()>
...
+
+    let ctx = LaunchContext {
+        launcher_id: launcher_id.to_string(),
+        working_dir: std::env::current_dir()?,
+        base_env: std::collections::HashMap::new(),
+    };
+
+    let overlay = launcher.env_overlay(&ctx).await?;
+
+    if dry_run {
```
gabe-l-hart: I love the dry run functionality! Reading closely, though, I'm wondering if we need to move this logic to the launcher itself. Right now, a `launcher` is expected to have a single shell command that runs, but that may not always be the case, and even with the current abstraction, there's no guarantee that `launcher.launch(args, &ctx)` actually runs the command returned by `launcher.validate_command()`.

A couple of other ideas to consider:

1. We could add `dry_run` to `LaunchContext` and let `launcher.launch` handle it, likely with some shared utility for `run_launcher_command` that takes a `LaunchContext` and handles the `dry_run` flag.
2. We could add a separate `.dry_run` abstract method to the `trait Launcher`. This would still allow individual launchers to define what a dry run is, but would lose the common-code reuse.

src/launchers/base.rs
gabe-l-hart: I actually meant that this should be at the top of the file with the rest of the `use` statements. I'm not super familiar with rust conventions, but in most languages, all imports/includes/use statements tend to live at the top of a file so that it's clear what that file depends on. Ideally, I'd also like to start enforcing a convention where we sort `use` statements into sections (`Standard`, `Third party`, `Local`) so it's clear which things come from the standard lib, third-party libs, or elsewhere in this project.

src/config/mod.rs
```diff
@@ -18,28 +18,17 @@ pub struct Config {
...
-#[derive(Debug, Clone, Serialize, Deserialize)]
+#[derive(Debug, Clone, Serialize, Deserialize, Default)]
 pub struct ModelConfig {
     pub model_id: String,
     pub provider_id: Option<String>,
     pub variant: Option<String>,
-    pub enabled: bool,
-}
-
-impl Default for ModelConfig {
```
gabe-l-hart, line 34: Why is this whole default removed? Is this implicit with the `#[derive(..., Default)]` above?

src/commands/launcher.rs
gabe-l-hart: Hm, I'm still not sure this is working. Here's what I'm seeing:

```sh
$ ./target/debug/granite-cli launcher setup claude

Setting up launcher: claude
Anthropic's Claude CLI tool
Default command: claude (leave command_path blank to use PATH lookup)
Instance name: : claude-enterprise
  Binary found: /Users/ghart/.local/bin/claude

Launcher 'claude-enterprise' configured successfully!
```

There was no pause to edit on `Default command: claude (leave command_path blank to use PATH lookup)`. I was also never prompted to enter `command_path` afterwards, so something isn't correctly recursing to the `schema` and then filling it.

src/commands/launcher.rs
gabe-l-hart: I'm still not sure this is fully fixed. I never saw any prompt based on the type-specific schema.

gabe-l-hart (changes requested):
Three things left:

1. I realized we need to rethink `dry_run` on launcher a bit
2. I'm still seeing issues with `launcher setup` not prompting for the sub-schema (will do some investigation myself too)
3. NIT about putting the `use` at the top of the file still needs handling

src/commands/launcher.rs
gabe-l-hart: Digging further into it, the issue seems to be that `is_promptable` is not catching the array type for `command_path` somehow. Still digging, but I think this is a bug in `prompt_for_schema`

gabe-l-hart: Ah, the issue is that the type itself is an `Array` of two elements to hold `String("string")` and `String("null")`

gabe-l-hart: Ok, fixed it! We needed some unwrapping logic in `prompt_for_schema`

src/launchers/base.rs
gabe-l-hart: Fixed and pushed

gabe-l-hart (approved):
Good to go now!

PR #35, opened 2026-08-04
Version command

gabe-l-hart:
## Description

This PR adds a new subcommand: `granite-cli version` that prints off a version string grounded in the git release version.


gabe-l-hart:
Looks like I may need to actually commit a `version.rs` stub then ignore future changes to it

Tanker2020 (commented):
The --version / -V flag was never wired up, recommend quick addition

#[command(version = version::version_string())]

gabe-l-hart:
>The --version / -V flag was never wired up, recommend quick addition
>
>#[command(version = version::version_string())]

I realized this, but then also realized that the `-v | --version` semantics are a bit ugly because it's not clear whether they should be a global flag (eg `-h | --help`) or just present on the top-level, and if it _was_ global, would we expect different semantics on subcommands somehow? All this led me to decide to just keep it simple and only have one way to print the version information (`granite-cli version`).

Issue #47, opened 2026-08-07
fix: Handle dangling references on `Remove`


gabe-l-hart:
## Description

While working on the implementation of #19, I realized we have an issue with how `Remove` effects downstream dependencies:

> NOTE: This actually raises a really important issue with the `Remove` operation. Right now, `Remove` is on a per-type basis, but that could easily result in stranding its downstream dependencies. I think there are two ways we could handle this: (1) cascading removal where removing a dependency of a configured instance is detected and the user is presented the option to either reconfigure that instance or also remove it, or (2) lazy rectification where dangling pointers are identified at usage time (instantiation time) and the user is offered the option to reconfigure at that point. I think (2) sounds a bit easier to implement, but (1) is probably tighter. For now, let's hold this as a future TODO (I'll make an issue), but depending on where this branch lands, we may come back and tackle it.

As a concrete example with the current state of `main`, if a `Model` depends on a `Provider` instance that is removed, at construction time, that `Model` won't have a valid/healthy provider (it will error out when constructing it).

afrittoli:
This is mostly implemented in spec 0024.
There are some follow-up issues identified in #123 and #128 which will help close this down completely.

PR #48, opened 2026-08-07
Capabilities core 19

### LINKED ISSUE #19: feat: Capability architecture

### LINKED ISSUE #52: feat: Add `ToolCalling` and `Thinking` model functions to model catalog
## Description

The `ModelFunction` enum supports both `ToolCalling` and `Thinking` which may be critical for certain capabilities (eg `AgentModel`). Currently, `models.yaml` is populated via `update-models` and doesn't support tagging language models with either `ToolCalling` or `Thinking`. This prevents `capabilities` from requiring a model with these functions.

gabe-l-hart:
## Description

Implements the Capability abstraction end-to-end: capabilities declare structured dependencies on models, providers, and external shell tools. Those dependencies are resolved generically at setup time against the live catalog/config (including the option to configure a brand-new instance on the spot) and the resolved capability produces launcher-ready Binding payloads at bind time. The first concrete capability, AgentModelCapability, surfaces a configured model's connection details (base URL, model name, endpoint path, auth, TLS) to a launcher for a requested ApiType.

A key design point: for now, a capability only ever depends on a model_id. The model's own configured provider is discovered transitively through the model (Model::provider()), rather than the capability separately tracking a provider_id that could point at a different, conflicting provider.

## Changes

- **Core capability traits (`src/capabilities/base.rs`)**: Capability, Dependency (Model/Provider/ExternalTool variants, each carrying config_key/requirement/resolved_id/required), and a `define_bindings!` macro that generates BindingType, BindingRequest, and Binding together so a new binding surface can't be added to one enum without the matching variant in the others.

- **Generic dependency resolution (src/dependency, src/capabilities/requirement.rs)**: `Requirement<U>/Configured<U>` traits and `dependency::resolve()` evaluate a capability's declared constraints against configured instances and the catalog, returning both what already satisfies the requirement and what catalog types could satisfy it if configured. Concrete constraint types: `ModelRequirement` (family, model_type, min_context_length, min_size, tags, supported_functions), `ProviderRequirement` (api_types OR, functions/endpoints AND), `ShellCommandRequirement` (PATH lookup).

- **First concrete capability (src/capabilities/agent_model.rs)**: `AgentModelCapability` / `AgentModelCapabilityConfig`, binding to `BindingType::AgentModel`.

- **Model → Provider chain**: `Model::provider()` gains a real default implementation that constructs the model's own configured provider (src/models/base.rs, codegen in build.rs, wired through ModelSource::from_config in src/models/mod.rs). commands/model.rs's setup()/pull() now reuse model.provider() instead of duplicating provider-lookup logic.

- **Generic, interactive capability setup (src/commands/capability.rs)**: two-phase wizard — prompt non-dependency config fields from the schema, then resolve each declared Dependency against the current catalog. Resolution always offers a "configure a new instance" option alongside any existing satisfying instances, drives the user through ModelCommands::setup/ProviderCommands::setup, and re-verifies the freshly configured instance actually satisfies the requirement before accepting it (bailing or warning otherwise, since the user could configure something incompatible).

- **Providers (ollama, openai-compatible, llama-cpp, lm-studio)**: added base_url()/api_key()/verify_ssl() accessors needed by AgentModelCapability::bind(), plus previously-missing Transcription endpoint support on ollama/llama-cpp.

- **Launchers**: `supported_capabilities()` changed from `Vec<String>` (capability ids) to `HashSet<BindingType>` (binding surfaces) — a launcher declares which binding shapes it understands, not which capability instances happen to exist. ClaudeLauncher now declares `BindingType::AgentModel` support. _DEFERRED_: Full support for actually configuring `AgentModel` into the `claude` executable via the environment.

- **Config**: CapabilityConfig gains a capability_type field (serialized as type) and its config moved from HashMap<String, String> to serde_json::Value, matching ModelConfig/ProviderConfig.

- **CLI**: capability setup now takes a catalog capability_type plus an optional --id nickname (was a single capability_id), so multiple named instances of the same capability type can be configured.

## Testing

- cargo build --all-targets, cargo test (372 passed / 0 failed), and cargo clippy --all-targets -- -D warnings all clean.

- New/updated unit tests across capabilities::agent_model, capabilities::mod, capabilities::requirement, commands::capability, commands::model, and models::mod, covering: dependency resolution narrowed by transitive provider support, the "configure a new instance" offer and post-configuration re-verification (via deliberately unsatisfiable requirements against the pure resolution helpers, to avoid driving live setup flows against real registries/network from unit tests), and bind()-time defensive checks.

- Manual: `granite-cli capability setup agent-model`
  - Verified path with existing model
  - Verified path with new model -> existing provider

## Related Issue(s)

Closes #19.
Closes #52.

Not addressed here (deliberately deferred, tracked separately): #47 — dangling capability references when a model/provider it points at is later removed.

Deferred support for `claude` env binding: https://github.com/ibm-granite-community/granite-cli/issues/49

## AI Usage

Full — built in collaboration with Claude Code (Sonnet 5), per this repo's AI-usage commit-trailer convention. Flow: an initial raw implementation from a detailed multi-part spec prompt, a round of structural review feedback (removing capability→capability peer dependencies, deduplicating BindingType/BindingRequest/Binding via macro, moving abstract binding types to base, making bind() a required trait method instead of a default no-op, ShellToolRequirement → ShellCommandRequirement, binding_types → supported_binding_types) applied by Claude, then follow-on work adding transitive model→provider resolution and, most recently, the "always offer to configure a new dependency instance" UX with post-configuration re-verification and its test coverage.


src/capabilities/base.rs
```diff
@@ -14,8 +113,22 @@ pub trait Capability: ConfigConstructable + Send + Sync {
     fn description(&self) -> &str;
     fn dependencies(&self) -> Vec<Dependency>;
 
+    /// Which binding surfaces this capability instance can fill.
+    fn binding_types(&self) -> HashSet<BindingType> {
```
Tanker2020: The PR description says `bind()` was explicitly made a required method (no default no-op) after design review. But `binding_types()` still has a default that returns an empty set — a capability that forgets to override it will silently advertise zero binding surfaces and never be matched by any launcher, with no compile error. Since `bind()` is already required, `binding_types()` should be too. Remove the default body here.

src/capabilities/agent_model.rs
```diff
@@ -0,0 +1,486 @@
...
+use crate::registry::ConfigConstructable;
+use async_trait::async_trait;
+use serde::{Deserialize, Serialize};
+use std::collections::HashSet;
+
+/*-- AgentModelCapabilityConfig ---------------------------------------------------*/
+
+#[derive(Debug, Clone, Serialize, Deserialize, Default, schemars::JsonSchema)]
+pub struct AgentModelCapabilityConfig {
+    #[serde(default)]
```
Tanker2020: `#[serde(default)]` on a `String` means a config missing `model_id` silently deserializes to `""`. The failure only surfaces at `bind()` time with the confusing message `"model '' not configured"`. Consider `Option<String>` so the absence is explicit and can be caught early, or at minimum validate at construction that the string is non-empty.

src/commands/capability.rs
```diff
@@ -94,76 +106,350 @@ impl CapabilityCommands {
...
+        if usable.is_empty() && configurable_types.is_empty() {
+            if required {
+                anyhow::bail!(
+                    "No configured model satisfies this capability's requirements yet, and none can be configured. Configure a compatible model and provider first."
+                );
             }
+            return Ok(None);
+        }
+
+        const CONFIGURE_NEW: &str = "Configure a new model...";
```
Tanker2020, line 292: This sentinel is pushed into the same `options` vec as real instance ids and compared by string equality. If a user ever names a model instance `"Configure a new model..."`, the configure-new branch fires incorrectly. Same issue on the provider side at line 392. A cleaner approach is to track the selection as an index offset or use a small enum/Option rather than a magic string mixed into the data list.

src/commands/capability.rs
```diff
@@ -94,76 +106,350 @@ impl CapabilityCommands {
...
+            let index = ctx
+                .ui
+                .select("Select a model type to configure:", &type_options, 0)?;
+            configurable_types[index]
+        };
+
+        ModelCommands::setup(ctx, model_type).await?;
+
+        let (usable_after, _) = Self::model_candidates(ctx, requirement);
+        if usable_after.contains(&model_type.to_string()) {
```
Tanker2020: After `ModelCommands::setup` returns, the re-check compares against `model_type` (the catalog type string, e.g. `"granite-3.1-8b-instruct"`). This works today only because model catalog keys double as their config keys. But the resolution path through `model_candidates` already returns configured instance ids — it would be more robust to just re-run `model_candidates` and check whether the newly configured entry appears in `usable_after` by taking `usable_after[0]` (or the diff between before and after) rather than re-constructing the id from `model_type`.

src/config/mod.rs
```diff
@@ -40,10 +40,22 @@ impl Default for ProviderConfig {
...
+    #[serde(rename = "type")]
+    pub capability_type: String,
+    pub config: serde_json::Value,
+}
+
+impl Default for CapabilityConfig {
+    fn default() -> Self {
+        Self {
+            capability_id: String::new(),
+            capability_type: String::new(),
```
Tanker2020: `Default` produces a `CapabilityConfig` with `capability_type: ""` — a value no registry will ever match, meaning any code that accidentally calls `default()` and serializes will silently write an invalid config record. Since there's no valid "empty" capability config, consider removing the `Default` impl entirely and requiring callers to supply a real `capability_type` at construction. If `Default` is needed for test helpers, a `#[cfg(test)]` impl scoped to tests would be safer.

src/commands/capability.rs
```diff
@@ -94,76 +106,350 @@ impl CapabilityCommands {
...
+                    requirement,
+                    required,
+                    ..
+                } => {
+                    if let Some(id) =
+                        Self::resolve_model_dependency(ctx, &requirement, required).await?
+                    {
+                        config
+                            .as_object_mut()
+                            .unwrap()
```
Tanker2020, line 213: `prompt_from_schema` returns a `serde_json::Value`. `.as_object_mut().unwrap()` will
panic at runtime if it ever returns anything other than `Value::Object` — e.g. on a
schema mismatch or a future change to that function. Replace both with:

    let obj = config.as_object_mut()
        .ok_or_else(|| anyhow::anyhow!("capability config is not a JSON object"))?;
    obj.insert(config_key, serde_json::Value::String(id));

Same fix needed at line 226 for the Provider branch.

src/models/mod.rs
```diff
@@ -12,6 +13,59 @@ pub static MODEL_REGISTRY: LazyLock<base::ModelFactory> = LazyLock::new(|| {
...
+                    .and_then(|pid| config.get_provider(pid))
+                {
+                    Some(provider_config) => {
+                        serde_json::json!({ "provider_config": provider_config })
+                    }
+                    None => serde_json::json!({}),
+                };
+                MODEL_REGISTRY
+                    .construct(&model_config.model_id, &cfg)
+                    .ok()
```
Tanker2020, line 56: `.ok()` silently discards any construction error — including the case where a model id
that was valid in an older catalog version no longer exists after an update. The user's
configured model just vanishes from `instances()` with no warning, and later gets a
confusing "model not configured" error from `bind()`.

Consider logging a warning via `alog_channel!` on the `None` branch:

    .filter_map(|model_config| {
        let result = MODEL_REGISTRY.construct(&model_config.model_id, &cfg);
        if result.is_err() {
            alog_channel!(MessageLevel::Warn, "Could not construct model '{}': {:?}", model_config.model_id, result);
        }
        result.ok().map(|model| (model_config.model_id.clone(), model))
    })

The same applies to `CapabilitySource::from_config` in `src/capabilities/mod.rs:27`.

build.rs
```diff
@@ -56,14 +56,20 @@ fn generate_model_struct(model: &YamlModel) -> String {
...
+        "pub struct {struct_name} {{ provider_config: Option<crate::config::ProviderConfig> }}\n\n"
+    ));
 
     // ConfigConstructable implementation
     s.push_str(&format!(
         "impl crate::registry::ConfigConstructable for {struct_name} {{\n"
     ));
-    s.push_str("    fn new(_cfg: &serde_json::Value) -> Self { Self {} }\n");
+    s.push_str("    fn new(cfg: &serde_json::Value) -> Self {\n");
+    s.push_str("        let provider_config = cfg.get(\"provider_config\").and_then(|v| serde_json::from_value(v.clone()).ok());\n");
```
Tanker2020: If the stored `ProviderConfig` YAML has a malformed `type` field, this silently produces
`provider_config = None`. `Model::provider()` then returns "model has no configured
provider" — identical to the legitimate "no provider associated" case, so there's no way
for the user to tell their on-disk config is corrupt.

Change to log the deserialization error:

    let provider_config = cfg.get("provider_config").and_then(|v| {
        serde_json::from_value(v.clone())
            .map_err(|e| eprintln!("warn: failed to deserialize provider_config: {e}"))
            .ok()
    });

(Using `alog_channel!` would be ideal, but build.rs codegen runs before the crate's
logging infrastructure is available — at minimum an `eprintln!` distinguishes the error
from the no-provider case.)

src/capabilities/base.rs
```diff
@@ -14,8 +113,22 @@ pub trait Capability: ConfigConstructable + Send + Sync {
...
+        HashSet::new()
+    }
+
+    /// Resolve a `BindingRequest` into a concrete `Binding`, looking up this
+    /// capability's model dependency from the given source. A model's own
+    /// provider is reached via `Model::provider()`.
+    async fn bind(
```
Tanker2020, line 122: The `bind()` doc says it looks up the model dependency "from the given source", but
`ModelSource` snapshots `ProviderConfig` at construction time. If `provider setup` is
called after the source was built, `model.provider()` returns a provider with stale
config (old URL, old API key) — silently, with no error.

This is safe in today's CLI where every command builds `ModelSource` fresh, but the
parameter type `&(dyn Configured<dyn Model> + Sync)` puts no constraint on lifetime.
Please add a doc comment, or rework this:

    /// # Lifetime note
    /// `models` must be constructed from the current `Config` immediately before
    /// calling `bind()`. `ModelSource` snapshots provider config at build time;
    /// a stale source will silently use outdated connection details.

src/capabilities/agent_model.rs
```diff
@@ -0,0 +1,486 @@
...
+    fn binding_types(&self) -> HashSet<BindingType> {
+        HashSet::from([BindingType::AgentModel])
+    }
+
+    async fn bind(
+        &self,
+        request: BindingRequest,
+        models: &(dyn Configured<dyn Model> + Sync),
+    ) -> anyhow::Result<Binding> {
+        let BindingRequest::AgentModel(AgentModelBindingRequest { api_type }) = request;
```
Tanker2020: This compiles today only because `BindingRequest` has exactly one variant. The moment
a second binding surface is added to `define_bindings!`, this line becomes a
non-exhaustive pattern compile error in every existing `Capability` impl — not a
graceful runtime bail.

Replace with an explicit match so future variants produce a useful runtime error
rather than a surprise compile failure across all implementors:

    let api_type = match request {
        BindingRequest::AgentModel(AgentModelBindingRequest { api_type }) => api_type,
        other => anyhow::bail!(
            "AgentModelCapability does not handle {:?} binding requests",
            other.binding_type()
        ),
    };

src/capabilities/agent_model.rs
```diff
@@ -0,0 +1,486 @@
...
+            model_name: model_id,
+            endpoint_path: endpoint.path().to_string(),
+            api_key: provider.api_key().cloned(),
+            verify_ssl: provider.verify_ssl(),
+        }))
+    }
+}
+
+impl HasCapabilityMetadata for AgentModelCapability {
+    fn metadata() -> CapabilityMetadata {
```
Tanker2020, line 130: `metadata()` hard-codes `dependencies` with `ModelRequirement::default()` (wildcard),
while the live `dependencies()` method builds a requirement narrowed to the configured
`function`. These two can diverge silently — a launcher consulting metadata will see
"accepts any model function" while the running instance may refuse anything that isn't
`Chat`.

Consider either: (a) having `metadata()` return the same wildcard it always has but add
a comment that instance-level `dependencies()` may narrow further, or (b) documenting
that `metadata()` represents the *type-level* superset intentionally, and adding a test
that asserts `metadata().dependencies` is a superset of `dependencies()` for any concrete
config.

src/capabilities/requirement.rs
```diff
@@ -0,0 +1,388 @@
...
+impl Requirement<dyn Provider> for ProviderRequirement {
+    fn admits_type(&self, metadata: &ProviderMetadata) -> bool {
+        self.admits(
+            &metadata.supported_api_types,
+            |f| metadata.default_function_endpoints.contains_key(f),
+            metadata.default_function_endpoints.values(),
+        )
+    }
+
+    fn admits_instance(&self, instance: &dyn Provider) -> bool {
```
Tanker2020, line 147: `instance.supports_function(f)` (the closure passed to `self.admits`) calls
`instance.function_endpoints()` internally on every function check — allocating a new
`HashMap` each time. `endpoints` was already computed on line 148 for the third argument.
Re-use it:

    self.admits(
        &instance.supported_api_types(),
        |f| endpoints.contains_key(f),   // use already-computed map
        endpoints.values(),
    )

src/config/mod.rs
Tanker2020, line 230: The on-disk filename stem and the in-memory map key are both the instance id, but
`capability_id` inside the YAML body is a third copy of the same string. If a file is
renamed on disk outside the CLI, `load_dir` will key it by the new filename while the
struct carries the old `capability_id` — a silent mismatch that causes `get_capability`
to find the entry but return stale ids.

Either drop `capability_id` from the YAML body and reconstruct it from the filename on
load (as the map key already is), or assert equality during `load_dir` and log a warning
on mismatch. The same issue applies to `provider_id` and `model_id`.

src/providers/ollama.rs
```diff
@@ -111,6 +111,11 @@ impl OllamaProvider {
             vec![ApiEndpoint::OpenAIEmbeddings, ApiEndpoint::OllamaEmbeddings],
```
Tanker2020, line 138: Ollama's OpenAI-compatibility layer does not currently expose `/v1/audio/transcriptions`.
Adding it here means `ProviderRequirement::admits_type` will match Ollama for
transcription capabilities and `bind()` will produce a binding to an endpoint that
returns 404 at runtime.

Please add a code comment with the tracking issue/version when this becomes supported,
or gate it behind a feature flag / runtime version check. Same concern applies to the
identical addition in `llamacpp.rs`.

src/capabilities/agent_model.rs
```diff
@@ -0,0 +1,486 @@
...
+        let err = cap
+            .bind(
+                BindingRequest::AgentModel(AgentModelBindingRequest {
+                    api_type: ApiType::OpenAI,
+                }),
+                &models,
+            )
+            .await
+            .unwrap_err();
+        assert!(err.to_string().contains("no"));
```
Tanker2020: This assertion is too broad — the word "no" appears in many possible error messages and
this test would pass even if the wrong error path fired. Use a more specific substring,
e.g.:

    assert!(err.to_string().contains("no OpenAI endpoint for Chat"));

or at minimum:

    assert!(err.to_string().contains("endpoint"));

Tanker2020 (changes requested):
Good architecture overall — the `define_bindings!` macro, two-phase wizard, and `Model::provider()` consolidation are all solid design choices. Test coverage is thorough on the happy and expected-failure paths.

Left inline comments on the issues I found. The ones I'd block on before merge:

1. **`capability.rs:211,226`** — `as_object_mut().unwrap()` can panic in production; needs an `?`-returning alternative.
2. **`capability.rs:322`** — post-setup re-verification uses the catalog type key as a proxy for the instance id. Works today by coincidence; breaks if that coincidence ever stops holding, and leaves `ctx.config` half-written on failure.
3. **`base.rs:117`** — `binding_types()` having a no-op default while `bind()` is required means a new capability impl can compile fine while advertising zero binding surfaces to every launcher.
4. **`models/mod.rs:42` / `build.rs:70`** — silent `.ok()` swallowing on construction means a corrupt on-disk provider config produces "model has no configured provider", indistinguishable from the legitimate no-provider case.

Everything else is noted inline at the relevant lines.

src/capabilities/base.rs
gabe-l-hart: Yikes, good catch. I flagged this on intermediate review, but I guess I didn't check that it got fixed very well.

src/capabilities/agent_model.rs
gabe-l-hart: Cool. I've added `serde_valid` to ensure `min_length = 1` and removed `serde(default)`

src/commands/capability.rs
gabe-l-hart: Good catch! We do want the magic string since it actually gets displayed to the user, but the bookkeeping was problematic since it was checking for a match by string value not index.

src/config/mod.rs
gabe-l-hart: Good call, this is probably true for a lot of `Default` implementations. We should probably audit this everywhere

gabe-l-hart: Created https://github.com/ibm-granite-community/granite-cli/issues/50 to track

src/commands/capability.rs
gabe-l-hart: Hmm. This is indeed a nasty trail. The `config` object is created via `prompt_from_schema` with the schema in question coming from the `CAPABILITY_REGISTRY.config_schema(capability_type)` call above. For this to ever fail, the capability registry would need to have a a capability registered that has a non-object schema which _should_ never happen (and isn't subject to random runtime permutations). That said, there's theoretically nothing preventing a user from adding a new `Capability` that defines a non-object schema which would trigger this. I'd rather catch that at PR/build time, so I think the right answer is probably to somehow assert that `config_schema` is an object-schema at build time somehow.

gabe-l-hart: This ended up opening up a bigger can of worms: What I _really_ want is to have a single source of truth for the specific `struct` a given impl of a given factory-type uses, and for the `schema_for!` of that struct to always be returned by `config_schema` so that `prompt_for_schema` can introspect and populate it. I think there's a much more elegant solution that involves expanding `ConfigConstructable` so that it requires the config type to be defined which would give us a compile-time binding to the correct config struct and allow these programmatic methods (`config_schema` and `default_config`) to be auto-derived based on this defined type.

gabe-l-hart: Ok, I did a pretty significant overhaul here and I think it's way cleaner now. I still don't love leaving this kind of dangerous code lying around (even though it's not structurally impossible to trigger the unhappy case), so I'll probably leave comments to indicate why it's safe.

src/models/mod.rs
gabe-l-hart: Fixed for all types

build.rs
gabe-l-hart: We actually can use `alog_channel!` since the generated code is all macro-included into `models/mod.rs` which has a `use_channel!` and the right `alog` imports

src/capabilities/base.rs
gabe-l-hart: Hmm, I think this actually points at a much bigger issue: The _instances_ created by a factory are (a) not guaranteed to be unique (you can construct the same instance multiple times) and (b) do not have reference semantics between each other. This probably needs to be addressed in a follow-up issue where the factory keeps an in-memory instance map and `construct` returns the pre-constructed instance if already constructed _by reference_. We would then need to ensure that all referential usage actually has proper reference semantics. I'll open a separate issue to track this one.

gabe-l-hart: https://github.com/ibm-granite-community/granite-cli/issues/51

src/capabilities/agent_model.rs
gabe-l-hart: Oof, good catch!

gabe-l-hart: I think it probably got factored out due to linting warnings with the `other` path being unreachable. I've added this with an `#[allow(unreachable)]` and a comment

src/capabilities/agent_model.rs
gabe-l-hart: This is actually incorrect on multiple fronts. The default and the instance should _both_ require both `Chat` and `ToolCalling`. This raises a bigger problem: Currently no models show `ToolCalling` (or `Thinking`) as supported functions. I've created #52 to track adding that to the model catalog update process.

src/config/mod.rs
gabe-l-hart: Hmm. I think asserting in `load_dir` is the cleanest. Users can do a lot to monkey with the config layout, but they _shouldn't_ and I think it's reasonable to just catch obvious problems.

src/capabilities/agent_model.rs
gabe-l-hart: I ended up doing this inline since the tests were broken without it

src/providers/ollama.rs
gabe-l-hart: That's not actually true! https://github.com/ollama/ollama/blob/main/server/routes.go#L1905. Both Ollama and llama.cpp have recently added this support (though I think you have to read the code to find it).

gabe-l-hart:
Ok, @Tanker2020 I think I got through all the comments! Thanks for the thorough review.

gabe-l-hart:
Moving ahead with merging now. Any follow up change requests can be added as follow ups

Issue #51, opened 2026-08-11
feat: Instance Reference Semantics


gabe-l-hart:
## Description

In https://github.com/ibm-granite-community/granite-cli/pull/48#discussion_r3743024990, @Tanker2020 identified a potential bug where a `Model` instance surfaces the `ProviderConfig` that it is built with _by value_, meaning that if the referenced provider changes between `Model` construction and `Capability` binding, the surfaced information could be out-of-date.

This reveals a bigger issue with the current factory/registry architecture: The constructed instances are not singletons within the factory AND the constructed structs that have logical references between factory-constructed types actually hold that information by value instead of reference. The goal of this issue is to rearchitect the registry so that instances are held as singletons within the factory (though a different factory instance could create a new factory-instance-singleton for eg testing). We then also need to audit the methods used to expose dependencies between types and ensure that they use reference semantics back to the singleton instance in the factory.

Tanker2020:
## Proposed Fix: Singleton-per-Factory-Instance Architecture

### Root Cause (Confirmed)

Every generated model struct stores a **cloned `ProviderConfig` by value** at construction time (`build.rs:62`, `build.rs:71`). `ModelSource::from_config` snapshots the provider config into the model instance JSON blob during construction, and from that moment the model is completely detached from the live `Config`. `Model::provider()` (`src/models/base.rs:171`) uses only that frozen snapshot.

The same snapshot-at-construction pattern is replicated across `ProviderSource`, `CapabilitySource`, and `LauncherSource` — all four hold `Vec<(String, Box<dyn Trait>)` with no connection back to `Config`.

### Why a Small Patch is Wrong

A minimal fix (move resolution to call time, keep `Box`, add a `provider_for()` method) would:
1. Break `Model::provider()` — a clean and useful trait method
2. Get **completely overwritten** when the singleton PR lands
3. Cause two PRs of churn for one architectural problem

### Proposed Fix: Singleton-per-Factory-Instance

**Core idea:** The `XxxSource` structs change from `Vec<(String, Box<dyn Trait>)>` to `Vec<(String, Arc<dyn Trait>)>`. The `define_factory!` macro gains a `construct_shared` method that returns `Arc<dyn Trait>`. `Configured::instances()` yields `Arc` refs rather than bare `&` refs.

This means:
- A model instance is the **canonical single copy** for a given configured id within a `ModelSource`
[...]

PR #54, opened 2026-08-13
Agent model binding

### LINKED ISSUE #49: feat: Implement `AgentModel` binding for `Claude`
## Description

Follow on to #48 . This is to implement the actual env binding so that a given model can be bound into `claude` for real usage.

gabe-l-hart:
## Description

This PR closes the loop on actually binding Capabilities to launchers with the concrete binding of the `AgentModelCapability` to `ClaudeLauncher`.

## Changes

- Remove the `models` argument from `Capability.bind`. For capabilities that need a specific model, they should store it at construct time so they can use it at bind time without needing to look it back up by ID
- Add multi-select to Ui abstraction for capability selection
- Construct `AgentModelCapability` with the actual `model` to be used in `bind`
  - **NOTE**: The current implementation of this is a bit hacky using `with_config` to extract the config for the global model config and construct it. We need some structural fixes here around the relationships between resource instances and the registries. We'll tackle doing this better in #51 
- Add a path to have a per-provider/variant name alias so that a given model can determine the correct ID string to use in the configured provider
- Bind the real model env overrides into `ClaudeLauncher` for end-to-end launch

## Testing

- Rigorous unit testing
- Spot testing of individual code paths

## Related Issue(s)

Closes #49 

Deferring factory refactors to #52 

## AI Usage

AI was used extensively for this PR. I used a large combination of agents and models. All commits were made by me after reviewing agent output and editing if needed.

<details>
<summary>git-ai-stats</summary>

```sh
╔══════════════════════════════════════════════════════════╗
║           GIT AI USAGE ANALYSIS                          ║
╚══════════════════════════════════════════════════════════╝

📊 COMMITS BY AGENT

--- Aggregate ---
Commits                        |      Count
---------------------------------------------
none                           |         23
OpenCode + Qwen3.6-35b         |         14
Claude + Sonnet 5              |         12
IBM Bob                        |          8
OpenCode + GraniteXX           |          3
OpenCode + Qwen3.6-35b, OpenCode + Granite |          1
---------------------------------------------
TOTAL                          |         61

📊 COMMITS BY USAGE TYPE

--- Aggregate ---
Commits                        |      Count
---------------------------------------------
none                           |         23
draft                          |          5
full                           |         33
---------------------------------------------
TOTAL                          |         61

📈 LINES OF CODE BY AGENT

--- Aggregate ---
Agent                     |    Commits |  Additions |  Deletions
------------------------------------------------------------
none                      |         23 |        279 |         96
OpenCode + Qwen3.6-35b    |         14 |        536 |        357
Claude + Sonnet 5         |         12 |       2575 |        573
IBM Bob                   |          8 |       1362 |         75
OpenCode + GraniteXX      |          3 |        229 |         17
OpenCode + Qwen3.6-35b, OpenCode + Granite |          1 |         11 |         19
------------------------------------------------------------
TOTAL                     |         61 |       4992 |       1137

📈 LINES OF CODE BY USAGE TYPE

--- Aggregate ---
Usage Type           |    Commits |  Additions |  Deletions
-------------------------------------------------------
none                 |         23 |        279 |         96
draft                |          5 |        162 |         38
full                 |         33 |       4551 |       1003
-------------------------------------------------------
TOTAL                |         61 |       4992 |       1137
```

</details>

### DIFF
--- a/src/capabilities/agent_model.rs
+++ b/src/capabilities/agent_model.rs
@@ -7,13 +7,13 @@ use crate::capabilities::base::{
     CapabilityMetadata, Dependency, HasCapabilityMetadata,
 };
 use crate::capabilities::requirement::ModelRequirement;
-use crate::dependency::Configured;
 use crate::models::{Model, ModelFunction};
 use crate::registry::ConfigConstructable;
 use async_trait::async_trait;
 use serde::{Deserialize, Serialize};
 use serde_valid::Validate;
 use std::collections::HashSet;
+use std::sync::Arc;
 
 /*-- AgentModelCapabilityConfig ---------------------------------------------------*/
 
@@ -27,6 +27,7 @@ pub struct AgentModelCapabilityConfig {
 
 pub struct AgentModelCapability {
     config: AgentModelCapabilityConfig,
+    model: Arc<dyn Model>,
 }
 
 impl ConfigConstructable for AgentModelCapability {
@@ -35,7 +36,19 @@ impl ConfigConstructable for AgentModelCapability {
     fn new(cfg: &serde_json::Value) -> Self {
         let config: AgentModelCapabilityConfig =
             serde_json::from_value(cfg.clone()).unwrap_or_default();
-        Self { config }
+        let model = crate::models::MODEL_REGISTRY
+            .construct(&config.model_id, &serde_json::json!({}))
+            .expect("model must be in registry");
+        Self {
+            config,
+            model: Arc::from(model),
+        }
+    }
+}
+
+impl AgentModelCapability {
+    pub fn configured_model_id(&self) -> &str {
+        &self.config.model_id
     }
 }
 
@@ -65,11 +78,7 @@ impl Capability for AgentModelCapability {
         HashSet::from([BindingType::AgentModel])
     }
 
-    async fn bind(
-        &self,
-        request: BindingRequest,
-        models: &(dyn Configured<dyn Model> + Sync),
-    ) -> anyhow::Result<Binding> {
+    async fn bind(&self, request: BindingRequest) -> anyhow::Result<Binding> {
         let api_type = match request {
             BindingRequest::AgentModel(AgentModelBindingRequest { api_type }) => api_type,
             #[allow(unreachable_patterns)] // Will remove once more variants are available
@@ -78,15 +87,10 @@ impl Capability for AgentModelCapability {
                 other.binding_type()
             ),
         };
-        let model_id = self.config.model_id.clone();
+        let model_id = &self.config.model_id;
 
-        let (_, model) = models
-            .instances()
-            .into_iter()
-            .find(|(id, _)| id == &model_id)
-            .ok_or_else(|| anyhow::anyhow!("model '{model_id}' not configured"))?;
-
-        let provider = model
+        let provider = self
+            .model
             .provider()
             .map_err(|e| anyhow::anyhow!("model '{model_id}' has no usable provider: {e}"))?;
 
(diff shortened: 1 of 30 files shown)


src/capabilities/agent_model.rs
```diff
@@ -27,15 +28,86 @@ pub struct AgentModelCapabilityConfig {
...
+}
+
+impl AgentModelCapability {
+    /// Construct from capability config, using the provided global config to
+    /// resolve the model's provider (so `model.provider()` works at bind time).
+    ///
+    /// `cfg` contains the capability's config (e.g. `{"model_id": "my-model"}`)
+    /// where `model_id` is the key into `global_config.models`.  The resolved
+    /// `ModelConfig` supplies the catalog model ID and the provider ID.
+    pub fn with_config(cfg: &serde_json::Value, global_config: &crate::config::Config) -> Self {
```
gabe-l-hart: This whole `with_config` architecture is less-than-ideal. It exists because the factory requires the config at `construct` time rather than itself having access to the config directly so that it can construct simply by ID. This will all be refactored during #51 , so for now, we'll keep this workaround.

src/capabilities/agent_model.rs
```diff
@@ -27,15 +28,86 @@ pub struct AgentModelCapabilityConfig {
...
+            configured_variant,
+        }
+    }
 }
 
 impl ConfigConstructable for AgentModelCapability {
     type Config = AgentModelCapabilityConfig;
 
     fn new(cfg: &serde_json::Value) -> Self {
+        // This path is used by CAPABILITY_REGISTRY.construct() which is
```
gabe-l-hart: This will need to be refactored out too

src/capabilities/mod.rs
```diff
@@ -29,10 +29,19 @@ impl CapabilitySource {
             .capabilities
             .values()
             .filter_map(|capability_config| {
-                let result = CAPABILITY_REGISTRY.construct(
-                    &capability_config.capability_type,
-                    &capability_config.config,
-                );
+                let result = match capability_config.capability_type.as_str() {
+                    "agent-model" => {
```
gabe-l-hart: This special-case will go away with the refactor for registry and config

src/capabilities/agent_model.rs
```diff
@@ -27,15 +28,66 @@ pub struct AgentModelCapabilityConfig {
 
 pub struct AgentModelCapability {
     config: AgentModelCapabilityConfig,
+    model: Arc<dyn Model>,
+    /// The raw `"format/precision"` string from `ModelConfig.variant`, if the
+    /// user configured a specific variant. Used at bind time to resolve the
+    /// provider-specific model alias.
+    configured_variant: Option<String>,
```
gabe-l-hart, line 35: This seems like it should be unnecessary. Right now, the Model only holds the provider config, but that seems wrong!

src/capabilities/agent_model.rs
```diff
@@ -27,15 +28,66 @@ pub struct AgentModelCapabilityConfig {
...
+        let model_cfg = global_config
+            .models
+            .get(&config.model_id)
+            .unwrap_or_else(|| {
+                panic!(
+                    "Configured model '{}' not found in config.models",
+                    config.model_id
+                )
+            });
+        let provider_cfg = model_cfg
```
gabe-l-hart, line 59: Ok, this is definitely not correct. The second arg to a model's `construct` should be a `NoConfig` (as currently written) NOT a `ProviderConfig`. I can see why this was done since we didn't have in-constructor dependent instantiation as a pattern, but this is not the right workaround. This pairs with code in the `build.rs` that extracts `"provider_config"` from the given json blob and saves that to the model's `provider_config` field.

PR #55, opened 2026-08-13
Feat/instance reference semantics

### LINKED ISSUE #51: feat: Instance Reference Semantics

Tanker2020:
## Description

A `Model` instance held its `ProviderConfig` **by value** — a snapshot baked in at construction time. If a user edited a provider between `ModelSource::from_config()` and a capability `bind()` call, the model silently surfaced stale provider data.

This PR rearchitects the factory/registry layer to fix that:

- **Model structs now store only a `provider_id`** (a string key) instead of a full `ProviderConfig` copy. Provider data is resolved fresh from the live config map at call time via a new `ModelSource::provider_for(model)` method.
- **The factory gains a memoised `construct_shared()` path** that returns `Arc<dyn Trait>` instead of `Box<dyn Trait>`. Calling it twice with identical inputs returns the same `Arc` (pointer equality) within a single factory instance lifetime — giving true singleton-per-factory semantics as described in the issue.
- All four source types (`ModelSource`, `ProviderSource`, `CapabilitySource`, `LauncherSource`) are updated uniformly to use `Arc` and `construct_shared`.

## Changes

**`src/registry/mod.rs`**
- Added `construct_arc()` to the internal `TraitMetadata_` vtable
- Added `construct_shared(&self, name, cfg) -> Result<Arc<dyn Trait>>` to every generated factory — memoised by `(type_name, cfg_json)` under a single `Mutex` lock using `entry().or_insert_with()` (eliminates the TOCTOU race a two-lock check-then-insert would have)
- Cache key uses the full serialised JSON string, not a hash, so there is no collision risk
- Added 3 new tests: same-`Arc` pointer equality, distinct configs → distinct `Arc`s, independent caches across factory instances

**`build.rs` + generated model structs**
- Model structs: `provider_config: Option<ProviderConfig>` → `provider_id: Option<String>`
- `Model` trait: `provider_config()` → `provider_id()`, and `provider()` default now returns an unconditional error directing callers to `provider_for()` — any stray call site fails loudly

**`src/models/mod.rs`**
- `ModelSource::constructed` changed from `Vec<(String, Box<dyn Model>)>` to `Vec<(String, Arc<dyn Model>)>`
- Added `provider_configs: HashMap<String, ProviderConfig>` field — a snapshot of `config.providers` captured at source construction time
- Added `ModelConfigured` trait (defined here, re-exported from `dependency` for backward-compatible import paths) with a single method `provider_for(&self, model) -> Result<Box<dyn Provider>>` that resolves from the live map on every call
- `Configured<dyn Model>::instances()` return type updated to `Vec<(String, Arc<dyn Model>)>`

**`src/dependency/mod.rs`**
- Removed the `ModelConfigured` trait definition (it referenced `crate::models` and `crate::providers`, creating a downward coupling from the generic foundation layer into the domain layers)
- Replaced with `pub use crate::models::ModelConfigured` so all existing import paths continue to work

**`src/capabilities/agent_model.rs`, `src/capabilities/base.rs`**
- `bind()` signature: `&dyn Configured<dyn Model>` → `&dyn ModelConfigured`
- Provider resolution: `model.provider()` → `models.provider_for(model.as_ref())`
- Test `FakeSource` updated to implement `ModelConfigured`; `provider_result` field moved from `FakeModel` to `FakeSource`

**`src/commands/model.rs`, `src/commands/capability.rs`**
- All 4 remaining `model.provider()` call sites replaced with `source.provider_for(m.as_ref())`

**`src/providers/mod.rs`, `src/capabilities/mod.rs`, `src/launchers/mod.rs`, `src/utils/ui/app.rs`**
- `ProviderSource`, `CapabilitySource`, `LauncherSource` — same `Box→Arc` / `construct→construct_shared` treatment as `ModelSource` for consistency

**`.gitignore`**
- Added `src/version.rs` — this file is auto-generated by `build.rs` into `OUT_DIR` and pulled in via `include!()`; the previously committed copy contained hardcoded stale values

## Testing

- `cargo test` — **377 tests pass, 0 fail** (374 pre-existing + 3 new memoisation tests)
- `cargo clippy -- -D warnings` — clean
- New regression test `model_source_reflects_updated_provider_config_on_rebuild` directly exercises the original bug: constructs two `ModelSource`s from the same config with the provider URL changed between them and asserts the second source returns the updated URL immediately

## Related Issue(s)

Closes #51

## AI Usage

This PR was developed with AI assistance (IBM Bob):
- Used AI to perform an initial code review of the first commit against the issue requirements, which identified the TOCTOU race, hash collision risk, `src/version.rs` VCS issue, missing memoisation tests, and the `ModelConfigured` layering concern in `dependency/mod.rs`
- All follow-on fixes in the second commit were implemented by AI based on that review
- All code was verified by the author; tests and clippy were run to confirm correctness


.gitignore
```diff
@@ -6,3 +6,6 @@ tmp
 .bob
 .claude
 .codegraph
+
+# Build-generated source (written to OUT_DIR by build.rs, included via include!())
```
gabe-l-hart: I don't think we want this in here? The `OUT_DIR` should be under `target` and already ignored

gabe-l-hart:
@Tanker2020 Thanks for putting this together! At this point, I've had to run too fast on the other features and this branch is out-of-date enough that I don't think it's worth salvaging as-is. I'm going to close for now and revisit if/when the lack of reference semantics and instance registry become a problem for a real user flow.

PR #68, opened 2026-08-18
Fix: Windows sluggish when navigating with arrow keys

### LINKED ISSUE #64: bug: TUI data not aligned
## Description

When using the interactive TUI, there are two issues I'm finding:

1. In the `Recommend` tab, the rows shuffle randomly when you navigate
2. In both the `Recommend` and `Models` tabs, hitting Enter to see the details of a model does not show the right detailed information for the given model

ByeolBaram:
## Description

Closes #64  

On Windows, key events are processed differently compared to Macs. This caused pressing a key once to sometimes jump twice. In addition, on Windows, `detect_hardware()` runs considerably slower than Metal device detection. A solution to this is to cache the `detect_hardware()` result once and not everytime a render occurs.

Edit: Second Commit

The rendered row IDs didn't match the provider IDs because of filtering and sorting, which is why there was a mismatch in information given when selecting a model in the models or recommended tab. The solution was to derive the IDs from the rendered rows.

## Changes

- Tweak to how key events are processed to accommodate Windows users
- Cache the detected hardware, so repeated testing doesn't slow down navigation, particularly for Windows
- Hardware shouldn't change during navigation anyway, so multiple hardware checks are a waste. Caching seems like the right call. Without all the hardware calls, maybe there will be minor speedup benefits for Mac users too?

Edit: Second Commit
- Derive model IDs from the rendered rows as opposed to the provider IDs, in order to match them up for selecting models

## Testing

- Manual Testing

## Related Issue(s)

N/A

## AI Usage

### DIFF
--- a/src/utils/ui/app.rs
+++ b/src/utils/ui/app.rs
@@ -250,28 +250,48 @@ impl App {
 
     fn filtered_ids(&self, query: &str) -> Vec<String> {
         let q = query.to_lowercase();
-        // Hardware has no selectable rows; Recommend uses its own row source
-        let mut ids: Vec<String> = match self.section {
-            Section::Models => MODEL_REGISTRY
-                .entries()
-                .keys()
-                .map(|k| k.to_string())
+        // IDs must be returned in the same order as the browse table renders
+        // them so that self.row correctly indexes the highlighted entry.
+        //
+        // Models: catalog_rows uses sort_enriched_rows (family/version/size),
+        //         not alphabetical — derive IDs from there.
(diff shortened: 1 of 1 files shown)


ByeolBaram:
This now also tackles issue #64 . The original commit I did fixed part 1 of that issue unintentionally without me knowing that was your issue. The second updated commit tackles part 2 of it.

gabe-l-hart (approved):
Looks great! Fixes the Recommend section nicely on Mac as well

PR #74, opened 2026-08-21
feat: Add Hermes launcher

aanokh:
## Description

Adds a launcher for the Hermes agent CLI (Nous Research), supporting both agent-model and MCP capability bindings.

Hermes reads all of its configuration (model endpoint, MCP servers, credentials) from a single home directory relocatable via HERMES_HOME. It has no additive config mechanism and its hermes mcp add CLI is interactive, so this launcher follows the same strategy as pi: it builds a granite-cli-owned Hermes home under GRANITE_CLI_HOME/launcher-state/, generates a fresh config.yaml from the bound model and MCP servers, symlinks everything else through from the user's real Hermes directory (so their credentials, SOUL.md, memories, skills, and plugins still apply), and points the child process at it with HERMES_HOME.

The generated model section uses provider: custom for an OpenAI-compatible endpoint. The API key is written as an ${ENV_VAR} reference that Hermes expands at load time, so the secret stays out of the generated file. The user's own config.yaml is never modified, their mcp_servers entries are preserved in the merge, and only the model section is replaced.
Session/log/cache directories stay local to the launcher state dir similar to pi's session store. The symlink exclusion list can be adjusted.

## Changes

- New src/launchers/hermes.rs: main launcher file modeled closely to pi.rs with extra capability for MCP entries
- src/launchers/mod.rs: registers the hermes launcher type
- src/commands/launcher.rs: extends the catalog test with the new launcher

## Testing

- cargo test: 520 passing including 48 new hermes tests
- cargo fmt --check and cargo clippy --all-targets clean
- end-to-end manual test on Hermes install with Ollama, verifying that model is switched, granite cli MCP is added, and user MCPs and other environment like SOUL file are preserved. Dry run notification works properly

## AI Usage

This PR was developed using IBM Bob (code verified and tested manually)

### DIFF
--- a/src/commands/launcher.rs
+++ b/src/commands/launcher.rs
@@ -428,7 +428,7 @@ mod tests {
     }
 
     #[test]
-    fn catalog_contains_claude_bob_pi_and_opencode() {
+    fn catalog_contains_claude_bob_pi_opencode_and_hermes() {
         let ctx = test_ctx();
         LauncherCommands::catalog(&ctx).unwrap();
         let tables = tables!(ctx);
@@ -437,6 +437,7 @@ mod tests {
         assert!(rows.iter().any(|r| r[0] == "bob"));
         assert!(rows.iter().any(|r| r[0] == "pi"));
         assert!(rows.iter().any(|r| r[0] == "opencode"));
+        assert!(rows.iter().any(|r| r[0] == "hermes"));
     }
 
     // -- list ------------------------------------------------------------------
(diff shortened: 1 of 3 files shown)


gabe-l-hart:
@aanokh Sorry we didn't coordinate this better! I ended up adding `hermes` as part of https://github.com/ibm-granite-community/granite-cli/pull/75. Closing for now, but please check out the `hermes` support on `main` and feel free to submit any fixes/improvements you find.

PR #88, opened 2026-08-26
Can now add multiple capabilities consecutively

### LINKED ISSUE #73: feat: Launcher setup multiple new capabilities
## Description

Currently, when running `granite-cli launcher setup <launcher>`, if you select `[x] Configure a new capability...`, you get taken into the capability setup wizard, but once done with that, the setup wizard ends. Instead, once that new capability is set up, you should be routed back to the capability selection view with the new capability pre-selected and the option to select yet another `[ ] Configure a new capability...` option. This should repeat until the user does not select `Configure a new capability`.

ByeolBaram:
## Description

Let's you setup multiple capabilities without starting over

## Changes

- Minor changes to how it gets looped, so that it loops back until "Configure a new capability..." is not selected

## Testing

- Manual Testing

## Related Issue(s)

Closes #73 

## AI Usage

### DIFF
--- a/src/commands/launcher.rs
+++ b/src/commands/launcher.rs
@@ -281,72 +281,81 @@ async fn select_capabilities(
         return Ok(vec![]);
     }
 
-    let source = CapabilitySource::from_config(&ctx.config);
+    // `enabled` accumulates the user's final selections across loop iterations.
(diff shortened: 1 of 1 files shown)


gabe-l-hart (approved):
Looks good to me! Tested and working locally

PR #89, opened 2026-08-27
fix: Enables empty input for optional string inputs

### LINKED ISSUE #84: bug: Allow empty "command_path" when configuring a launcher
## Description

When running `granite-cli launcher setup <name>`, the `prompt_for_schema` wizard correctly recurses into the underlying config and asks for `command_path`, but if you hit `Enter` without typing one, it does nothing. Instead, it should accept that this is a "none" response since the field is an `Option<String>` and move on.

afrittoli:
## Description

Fixes #84: `launcher setup` silently ignored a blank answer for optional string fields (e.g. `command_path`) instead of treating it as "unset" — pressing Enter on an empty prompt looked like it did nothing. 

Root cause was two layers deep: the prompt wizard didn't know a blank answer was meaningful for `Option<String>` fields, *and* the underlying `dialoguer` input rejected empty submissions outright, so even after teaching the wizard what to do with a blank answer, it could never actually receive one.

## Changes

Two commits, in dependency order:

1. **`fix: allow blank in Ui::text for optional fields`** — adds an `allow_empty: bool` parameter to `Ui::text()` so callers can selectively decided whether to allow empty or not. Will be used by `prompt_string`.

2. **`fix: handle empty input for Option<String> fields`** — `prompt_string()`/`prompt_value()` detect `Option<T>` from the JSON schema (`is_optional_field()`), pass `true` for `allow_empty` only in that case, and map a blank answer to `Value::Null` instead of `Some("")`. Also removes the narrower `"" -> Null` normalization that used to live in `commands/launcher.rs`, scoped only to `command_path` — no longer needed once the wizard handles it generically for any optional field.

## Testing

- `cargo test` (638/638) and `cargo clippy --all-targets --all-features -- -D warnings` clean at **both** commits individually, not just the final state.
- Manually reproduced end-to-end via `cargo run -- launcher setup claude`: a blank `command_path` now falls back to PATH discovery instead of failing binary validation.

## Related Issue(s)

Fixes: #84

## AI Usage

Draft (Claude + Sonnet 5) for commit 1
Draft (Bob) for commit 2

### DIFF
--- a/src/commands/capability.rs
+++ b/src/commands/capability.rs
@@ -146,7 +146,8 @@ impl CapabilityCommands {
 
         let instance_id = match instance_id {
             Some(id) => id.to_string(),
-            None => ctx.ui.text("Instance name: ", capability_type)?,
+            // Becomes a config map key -- must not be empty.
+            None => ctx.ui.text("Instance name: ", capability_type, false)?,
         };
 
         let existing_config = ctx.config.get_capability(&instance_id);
@@ -447,7 +448,10 @@ impl CapabilityCommands {
             configurable_types[index]
         };
 
-        let nickname = ctx.ui.text("Name this provider instance", provider_type)?;
+        // Becomes a config map key -- must not be empty.
+        let nickname = ctx
+            .ui
+            .text("Name this provider instance", provider_type, false)?;
         ProviderCommands::setup(ctx, provider_type, Some(&nickname)).await?;
 
         let (existing_after, _) = Self::provider_candidates(ctx, requirement);
--- a/src/commands/capability.rs
+++ b/src/commands/capability.rs
@@ -146,8 +146,7 @@ impl CapabilityCommands {
 
         let instance_id = match instance_id {
             Some(id) => id.to_string(),
-            // Becomes a config map key -- must not be empty.
-            None => ctx.ui.text("Instance name: ", capability_type, false)?,
+            None => ctx.ui.text("Instance name: ", capability_type)?,
         };
 
         let existing_config = ctx.config.get_capability(&instance_id);
@@ -448,10 +447,7 @@ impl CapabilityCommands {
             configurable_types[index]
         };
 
-        // Becomes a config map key -- must not be empty.
-        let nickname = ctx
-            .ui
-            .text("Name this provider instance", provider_type, false)?;
+        let nickname = ctx.ui.text("Name this provider instance", provider_type)?;
         ProviderCommands::setup(ctx, provider_type, Some(&nickname)).await?;
 
         let (existing_after, _) = Self::provider_candidates(ctx, requirement);
--- a/src/utils/ui/base.rs
+++ b/src/utils/ui/base.rs
@@ -220,12 +220,27 @@ pub trait Ui: Send + Sync + Any {
     /// Ask for a line of free text, pre-filled with `default`.
     ///
     /// `allow_empty` controls whether submitting a blank line (Enter with no
-    /// input) is accepted: dialoguer rejects an empty submission by default,
-    /// re-prompting silently instead of returning "" -- from the terminal
-    /// this looks like Enter does nothing. Every caller currently passes
-    /// `false`, preserving that behavior; a follow-up change teaches
-    /// `prompt_string()` in `prompt.rs` to pass `true` for `Option<String>`
-    /// fields, where a blank answer is a legitimate "unset".
+    /// input) is accepted:
+    ///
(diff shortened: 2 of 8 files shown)


src/utils/ui/base.rs
```diff
@@ -218,10 +218,34 @@ pub trait Ui: Send + Sync + Any {
...
+    ///
+    /// dialoguer rejects an empty submission by default, re-prompting
+    /// silently instead of returning "" -- from the terminal this looks
+    /// like Enter does nothing. That's the *right* behavior for most
+    /// callers here: instance names (`ctx.ui.text("Instance name: ", ..)`
+    /// in `commands/{provider,launcher,capability}.rs`) become config map
+    /// keys and must not be empty, and `prompt_number` in `prompt.rs`
+    /// would otherwise hand a blank string to `str::parse`.
+    ///
+    /// The one caller that needs `allow_empty = true` is `prompt_string()`
```
gabe-l-hart: This comment is probably a bit too verbose. No need to change just for this, but if we make any other changes, let's clean this part up a bit.

src/utils/ui/base.rs
```diff
@@ -218,10 +218,34 @@ pub trait Ui: Send + Sync + Any {
...
+    /// `Option<String>`: there, an empty submission is a legitimate answer
+    /// meaning "unset" (e.g. an unset launcher `command_path`, meant to
+    /// fall back to PATH discovery at `validate_command()` time), and
+    /// `prompt_string` already knows -- from the JSON schema -- whether
+    /// that's the case, so it's the right place to make this call, not the
+    /// default for every text prompt in the app. See issue #84: the
+    /// previous commit taught `prompt_string` to map an empty optional
+    /// string to `None`, but it had no effect until this commit let an
+    /// empty answer reach it in the first place.
+    fn text(&self, prompt: &str, default: &str, allow_empty: bool) -> anyhow::Result<String> {
```
gabe-l-hart: It seems like the other option, rather than adding a new argument, would be to add a new interface method for `text_optional` and invoke that conditionally in `prompt_for_schema` based on the `nullable` value in the schema. This would be less invasive and probably more sustainable in the long run. Not a strong opinion, but I think I would like that path better as it would require less plumbing between the source of truth (the `Option<String>` in the config struct) and the actual prompting.

src/utils/ui/prompt.rs
```diff
@@ -301,8 +311,15 @@ fn prompt_string(
         };
         Ok(Value::String(value))
     } else {
-        let entered = ui.text(&prompt, &default_str)?;
-        Ok(Value::String(entered))
+        // Only Option<String> fields may submit a blank line; see the
```
gabe-l-hart: I think we should be able to determine whether the string is optional in the schema based on the `nullable`, but it depends on what `schemars` does here. Probably worth investigating to avoid needing to plumb a new arg through.

gabe-l-hart (commented):
A few thoughts on how to make it less invasive. Worth a little investigation, but not a blocker. The fix works as is, so if it's hard to do the clean up, we can just merge this.

src/utils/ui/prompt.rs
afrittoli: I thought it would be safer to check the node *before* it went through `resolve_ref`, but it's still possible to detect the optional case after it, as long as what happens before does not change.

src/utils/ui/base.rs
afrittoli: Fixed

src/utils/ui/base.rs
afrittoli: added text_optional

src/utils/ui/prompt.rs
afrittoli: I added some extra unit tests to clarify what is_optional_field expects, and simplified its implementation to what is actually used today

src/commands/capability.rs
```diff
@@ -146,6 +146,7 @@ impl CapabilityCommands {
 
         let instance_id = match instance_id {
             Some(id) => id.to_string(),
+            // Becomes a config map key -- must not be empty.
```
gabe-l-hart: I think we can revert these comment changes now and just limit the number of files touched

gabe-l-hart (changes requested):
Looks great! Just some dead comments to remove now, then ready to merge

src/commands/capability.rs
afrittoli: Done - I forced with lease, didn't need an extra commit just for this.

gabe-l-hart (approved):
Ship it!

PR #91, opened 2026-08-27
fix: Skip capabilities whose model was removed

### LINKED ISSUE #90: bug: launcher/provider/capability setup panics when a configured capability references a removed model
## Description

`granite-cli launcher setup <type>` (and any other command that builds a `CapabilitySource::from_config`) crashes with a panic instead of returning an error when *any* configured capability references a model that's no longer configured -- e.g. after `model remove <id>` on a model some capability still depends on.

The panic isn't scoped to the capability actually being used: `select_capabilities()` in `commands/launcher.rs` eagerly constructs *every* capability in config via `CapabilitySource::from_config()` before it even presents the multi-select, so one stale reference anywhere in config takes down the whole command.

## Steps to reproduce

1. `granite-cli model setup granite-4.2-8b`
2. `granite-cli capability setup agent-model` and use the `granite-4.2-8b` model
3. `granite-cli model remove granite-4.2-8b`
4. `granite-cli launcher setup claude`

## Expected behavior

Setup either skips the broken capability with a warning -- matching how `LauncherSource`/`ProviderSource::from_config` already skip and warn on an unknown *type* string rather than failing hard -- or fails with a normal `Result`-based error naming the broken capability and its missing model.

## Actual behavior

The process panics and exits:

```
thread 'main' panicked at src/models/base.rs:238:17:
Configured model '<model-id>' not found or could not be constructed
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
```

afrittoli:
## Description

When building a launcher, if any capability had a dangling reference, panic was triggered so no new launcher could be created until the dangling reference was fixed. This is not a good user experience and not in line with the behavior of the CLI in other parts of the user experience.

This PR fixes this by removing the panic, replacing with an error which is stored and used to filter our unhealthy capabilities in the UI.

## Changes

`ConfiguredModel::resolve()`panicked when a capability's model_id no longer existed in config (e.g. after `model remove`), crashing any command that eagerly constructs capabilities.
Since `ConfigConstructable::new` is infallible by trait signature, the panic bypassed `CapabilitySource::from_config's existing skip-and-warn handling for construction failures.

`resolve()` now returns a Result instead of panicking; the five model-backed capabilities store it and surface the failure through a new `Capability::is_healthy()` check (used by
`CapabilitySource::from_config` to skip and warn, matching how an unknown capability_type is already handled) and through bind()'s existing Result.

## Testing

1. Root-cause level (src/models/base.rs)
- resolve_returns_named_err_for_unconfigured_model_id — asserts ConfiguredModel::resolve() returns Err naming the model id instead of panicking, for a model that isn't in config.

2. Capability level (src/capabilities/agent_model.rs)
- new_does_not_panic_when_model_id_is_unconfigured — constructs a real AgentModelCapability (via ConfigConstructable::new, not the test double) with a dangling model_id and checks is_healthy() reports the error.
- bind_fails_with_named_error_when_model_id_is_unconfigured — same setup, but calls bind() and asserts it returns a proper anyhow::Result error naming the model, not a panic.

3. Source/registry level (src/capabilities/mod.rs)
- capability_source_skips_capability_referencing_a_removed_model — CapabilitySource::from_config() with one capability pointing at a missing model → instances() comes back empty (skipped, not crashed).
- capability_source_skips_broken_entries_but_keeps_healthy_ones — same, but alongy, to confirm only the broken one gets dropped, not everything.
## Related Issue(s)

Fixes: #90 

## AI Usage

AI-Usage: draft (Claude Sonnet 5)

### DIFF
--- a/src/capabilities/agent_model.rs
+++ b/src/capabilities/agent_model.rs
@@ -28,7 +28,11 @@ pub struct AgentModelCapabilityConfig {
 pub struct AgentModelCapability {
     instance_id: String,
     config: AgentModelCapabilityConfig,
-    configured_model: ConfiguredModel,
+    /// `Err` when `config.model_id` doesn't resolve (e.g. the model was
+    /// removed after this capability was configured). Construction stays
+    /// infallible per `ConfigConstructable`; the error is surfaced at
+    /// `bind()` time and via `Capability::is_healthy`.
+    configured_model: Result<ConfiguredModel, String>,
 }
 
 impl ConfigConstructable for AgentModelCapability {
@@ -107,8 +111,12 @@ impl Capability for AgentModelCapability {
             ),
         };
         let model_id = &self.config.model_id;
+        let configured_model = self
+            .configured_model
+            .as_ref()
+            .map_err(|e| anyhow::anyhow!(e.clone()))?;
 
-        let (provider, endpoint, model_name) = self.configured_model.resolve_provider_endpoint(
+        let (provider, endpoint, model_name) = configured_model.resolve_provider_endpoint(
             model_id,
             api_type.clone(),
             ModelFunction::Chat,
@@ -123,8 +131,15 @@ impl Capability for AgentModelCapability {
             endpoint_path: endpoint.path().to_string(),
             api_key: provider.api_key().cloned(),
             verify_ssl: provider.verify_ssl(),
-            context_length: Some(self.configured_model.model.context_length()),
+            context_length: Some(configured_model.model.context_length()),
         }))
+    }
+
+    fn is_healthy(&self) -> Result<(), String> {
+        self.configured_model
+            .as_ref()
+            .map(|_| ())
+            .map_err(Clone::clone)
     }
 }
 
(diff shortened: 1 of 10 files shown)


gabe-l-hart:
I think it might be worth tackling the full #47 here as well. This dangling model case is a special case of any dangling reference in config. I think we should think carefully about how we handle this holistically. My gut is the following:

1. Config is read from disk exactly once (at process start). At startup, we validate all references and resolve dangling references proactively (user has the choice to re-instantiate the missing ref, or delete the configured instance that depends on the dangling type).
2. Once in memory, the authoritative source-of-truth is the in-memory version, so if something messes with config on disk while a process is live, it is not loaded or updated in-process
3. If config is _written_ from a process, we would either need some kind of conflict resolution or we'd need to simply overwrite whole-cloth with the in-memory version.

I think the _best_ solution would be conflict resolution on read and write, but the write resolution might be tricky and we would still have loads of race conditions (eg further conflicts written while conflicts are being resolved). We could consider using some kind of file-based lock system during read and write to mitigate that.

gabe-l-hart:
Closing in favor of design from #106

PR #97, opened 2026-08-27
Added Docling Capability

### LINKED ISSUE #8: feat: Docling Capability
## Description

Add a `capability` that will install/run https://github.com/docling-project/docling-mcp enabled with `granite-docling` models.

ByeolBaram:
## Description

- Adds Docling as a Capability

## Changes

- Adds the ability for a docling to be used in order to convert documents

## Testing

- Manual Testing
- AI Testing

## Related Issue(s)

Closes #8 

## AI Usage

### DIFF
--- a/resources/models.yaml
+++ b/resources/models.yaml
@@ -2537,6 +2537,7 @@
   supported_functions:
     - Chat
     - ImageUnderstanding
+    - DocumentConversion
   variants:
     - format: GGUF
       precision: BF16
@@ -2582,6 +2583,7 @@
   supported_functions:
     - Chat
     - ImageUnderstanding
+    - DocumentConversion
   variants:
     - format: safetensors
       precision: null
--- a/src/capabilities/docling_mcp/mod.rs
+++ b/src/capabilities/docling_mcp/mod.rs
@@ -1,0 +1,303 @@
+//! `DoclingMCPCapability`: installs and runs the
+//! [docling-mcp](https://github.com/docling-project/docling-mcp) server via
+//! `uvx`, wiring it to a configured granite-docling model's provider endpoint
+//! so document-conversion tools are available to any launched coding agent.
+//!
+//! The capability resolves the model's provider at `bind()` time and injects
+//! its base URL and API key as `DOCLING_MCP_SERVICE_URL` /
+//! `DOCLING_MCP_SERVICE_API_KEY` environment variables, enabling docling-mcp's
+//! remote mode against the serving endpoint. The model must advertise the
+//! `DocumentConversion` function (only `granite-docling-*` models carry this
+//! tag today).
+//!
+//! `bind()` returns a `McpBinding::Stdio` whose `command` is `uvx` — the
+//! downstream launcher passes that directly to the tool it is wrapping,
+//! which then spawns the server process.
+
+use crate::capabilities::base::{
+    Binding, BindingRequest, BindingType, Capability, CapabilityMetadata, Dependency,
+    HasCapabilityMetadata, LaunchContext, McpBinding, McpBindingRequest, McpTransportKind,
+};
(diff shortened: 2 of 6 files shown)


ByeolBaram:
@gabe-l-hart The last commit, I'm a little unsure on. To pass the cargo test, it wanted me to configure it to allow it to work even if there isn't a recommended model. I'm not sure if you want to accept that setup process though. Take a look and let me know what you think.

resources/models.yaml
```diff
@@ -2537,6 +2537,7 @@
   supported_functions:
     - Chat
     - ImageUnderstanding
+    - DocumentConversion
```
gabe-l-hart, line 2540: We need to make this change in the `.agents/skills/update-models` scripts rather than manually. This will require some special-case logic, but there's already precedent for that with Guardian that appears like a normal chat model. While you're in there, I'd like to remove `ImageUnderstanding` from these models since these models currently get incorrectly identified as valid general-purpose VLMs which they aren't.

src/launchers/shared/mcp_cli.rs
```diff
@@ -29,6 +33,9 @@ pub(crate) fn register_mcp_server(
     ctx: &LaunchContext,
     ui: &dyn Ui,
 ) -> anyhow::Result<()> {
+    // Best-effort: ignore failure — the server simply may not exist yet.
```
gabe-l-hart, line 36: Is this an attempt to clean up from previous failures?

src/models/base.rs
```diff
@@ -23,6 +23,8 @@ pub enum ModelFunction {
     ImageUnderstanding,
     /// Detect harms
     Guardian,
+    /// Document conversion and processing (e.g. PDF to structured formats)
+    DocumentConversion,
```
gabe-l-hart, line 27: Good call having this as a separate ModelFunction. This is fundamentally different than `ImageUnderstanding` in its usage.

src/capabilities/docling_mcp/mod.rs
```diff
@@ -0,0 +1,305 @@
...
+            ModelFunction::Chat,
+        )?;
+
+        // Build the service URL: strip the trailing operation path so docling-mcp
+        // sees the API root (e.g. `http://localhost:11434/v1`), the same
+        // convention used by VisionMCPCapability's `vlm_base_url`.
+        let service_url = service_base_url(provider.base_url(), endpoint.path());
+
+        let mut env = std::collections::HashMap::new();
+        env.insert("DOCLING_MCP_SERVICE_URL".to_string(), service_url);
```
gabe-l-hart, line 160: I don't think this works as written. The `DOCLING_MCP_SERVICE_URL` is expecting to hit an instance of the `Docling Serve` server (or the public SaaS service). Routing this to Ollama will just hit 404s since it's a totally different server.

This is the fundamental problem that kept me from implementing this earlier: the `docling-mcp` package is capable of running the default conversion pipeline internally (which itself includes several small models, so should not be run in `stdio` to avoid bootup costs). The problem is that it doesn't expose any configuration surface for the pipeline details, particularly enabling the VLM pipeline and setting the model. The only way to accomplish this is to hit an external `docling-serve` instance. `docling-serve` _does_ allow you to configure a remote VLM for conversion, so the right path forward is to boot an instance of `docling-serve` as a session-scoped server routing to the configured provider for the VLM, then do what you've done here and bind `docling-mcp` as `stdio` and simply shuttle the requests to the `docling-serve` instance.

gabe-l-hart (changes requested):
Two key changes needed here:

1. We should only ever edit `models.yaml` through changes in the `update-models` skill scripts
2. The core implementation is missing `docling-serve` which is needed to take advantage of a hosted docling model running in a configured provider.

PR #101, opened 2026-08-28
Catalog list

### LINKED ISSUE #4: feat: Catalog vs List in interactive TUI
## Description

Currently, the interactive TUI in `app.rs` only displays the catalog view for each type. We need to be able to show the full catalog as well as the set of configured instances.

### LINKED ISSUE #5: feat: Setup in TUI
## Dependencies

https://github.com/ibm-granite-community/granite-cli/issues/4

## Description

We need the ability to run the equivalent of `granite-cli <type> setup <type_id>` within the interactive TUI in `app.rs`.

ByeolBaram:
## Description

Adds the capability to see configured instances and setup instances within the TUI

## Changes

- Added a check mark next to configured instances and information to their details panel
- Added ability to configure instances and setups from within the TUI
- Changed details to [d] and [enter] now begins the setup process

## Testing

- Manual Testing
- AI Testing

## Related Issue(s)

Closes #4 
Closes #5 

## AI Usage

### DIFF
--- a/src/capabilities/agent_model.rs
+++ b/src/capabilities/agent_model.rs
@@ -109,10 +109,7 @@ impl Capability for AgentModelCapability {
         let model_id = &self.config.model_id;
 
         let configured_model = self.configured_model.as_ref().ok_or_else(|| {
-            anyhow::anyhow!(
-                "Model '{}' is not configured or could not be resolved",
-                model_id
-            )
+            anyhow::anyhow!("Model '{model_id}' is not configured or could not be resolved")
         })?;
         let (provider, endpoint, model_name) = configured_model.resolve_provider_endpoint(
             model_id,
--- a/src/capabilities/agent_model.rs
+++ b/src/capabilities/agent_model.rs
@@ -28,7 +28,7 @@ pub struct AgentModelCapabilityConfig {
 pub struct AgentModelCapability {
     instance_id: String,
     config: AgentModelCapabilityConfig,
-    configured_model: Option<ConfiguredModel>,
+    configured_model: ConfiguredModel,
 }
 
 impl ConfigConstructable for AgentModelCapability {
@@ -50,7 +50,7 @@ impl ConfigConstructable for AgentModelCapability {
     ) -> Self {
         let config: AgentModelCapabilityConfig =
             serde_json::from_value(cfg.clone()).unwrap_or_default();
-        let configured_model = ConfiguredModel::resolve(&config.model_id, global_config).ok();
+        let configured_model = ConfiguredModel::resolve(&config.model_id, global_config);
         Self {
             instance_id: instance_id.to_string(),
             config,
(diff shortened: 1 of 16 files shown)


gabe-l-hart (changes requested):
This is really slick! It's a great improvement on the setup experience. Three features I'd like to add before fully reviewing:

1. A column in the catalog table showing configured instance names for instances that are already configured
2. When selecting a type to view details on from the catalog table, before the bottom panel starts the setup dialog, it should allow you to select an instance to configure from the existing configured instances with an always-present option for "New instance"
3. A toggle to show _only_ configured types in the overall view that defaults to true if there are any configured instances and false otherwise

gabe-l-hart:
Also, I bet some of the changes I just made in #102 might break some of the config flow for models

ByeolBaram:
@gabe-l-hart I made the changes you wanted. Let me know how it looks, and if you want anything else changed!

src/capabilities/vision_mcp/mod.rs
```diff
@@ -80,7 +80,7 @@ impl Default for VisionMCPCapabilityConfig {
 pub struct VisionMCPCapability {
     instance_id: String,
     config: VisionMCPCapabilityConfig,
-    configured_model: ConfiguredModel,
+    configured_model: Option<ConfiguredModel>,
```
gabe-l-hart: Hm, I'm not clear why this needed to become optional. The capability won't work without a model

src/capabilities/agent_model.rs
```diff
@@ -28,7 +28,7 @@ pub struct AgentModelCapabilityConfig {
 pub struct AgentModelCapability {
     instance_id: String,
     config: AgentModelCapabilityConfig,
-    configured_model: ConfiguredModel,
+    configured_model: Option<ConfiguredModel>,
```
gabe-l-hart: Looks like all configured models became optional in capabilities. I'd prefer to avoid that if it's not strictly necessary

src/commands/setup.rs
```diff
@@ -1416,14 +1416,31 @@ impl SetupCommands {
 
         // Configure capabilities
         for cap_type in selected_caps {
+            let model_id = Self::find_model_for_capability(cap_type, selected_models);
+
+            // Skip capabilities that have a required model dependency but no
```
gabe-l-hart, line 1421: Hmm, I'm guessing this is related to all the optional parts?

src/models/base.rs
```diff
@@ -245,24 +245,24 @@ impl ConfiguredModel {
...
             .get(model_id)
             .and_then(|mc| mc.variant.clone());
         let mut source = crate::models::ModelSource::from_config(global_config);
         let model = source
             .take(model_id, configured_variant.as_deref())
-            .unwrap_or_else(|| {
-                panic!("Configured model '{model_id}' not found or could not be constructed")
-            });
-        Self {
+            .ok_or_else(|| {
```
gabe-l-hart: I'm guessing this is the reason for the `Option<...>`

gabe-l-hart (commented):
I pushed one commit (plus a `fmt`) to change the labels on the catalog toggle. The functionality looks good, but I'd prefer to avoid conflating this TUI change with changes to the optional status of models in `Capabilities`. Was there something that made that change necessary?

src/commands/model.rs
```diff
@@ -824,11 +824,9 @@ impl ModelCommands {
         resolution: &dependency::Resolution,
     ) -> Result<Option<String>> {
         if resolution.is_unsatisfiable() {
-            ctx.ui
-                .info("\nNo provider supports this variant's format/precision yet.");
-            ctx.ui
-                .info("Configure a provider later, then set its id on this model.");
-            return Ok(None);
+            anyhow::bail!(
```
gabe-l-hart, line 827: Properly bailing here is what allows us to avoid the `Option<...>` change everywhere

gabe-l-hart (approved):
I think this is good to go now!

Issue #58, opened 2026-08-14
refactor: introduce a global instance registry to replace global_config threading and baked-in dependencies


Tanker2020:
## Problem

There are two related structural issues that make the wiring between components increasingly painful to extend.

### 1. `global_config` threading through `ConfigConstructable::new`

`ConfigConstructable::new` currently has the signature:

```rust
fn new(cfg: &serde_json::Value, global_config: &Config) -> Self
```

The `global_config` arg exists only because `AgentModelCapability` needs to resolve `model_id → ModelConfig → provider_id → ProviderConfig` at construction time. Every other type (`OllamaProvider`, `BobLauncher`, etc.) ignores it entirely but is forced to accept and thread it anyway. This will get worse as more cross-component types are added.

### 2. The `.config` blob loses the typed outer fields

`ProviderConfig`, `CapabilityConfig`, and `LauncherConfig` each carry a `config: serde_json::Value` inner blob **plus** typed outer fields (`provider_id`, `provider_type`, `enabled_capabilities`). When only the blob is passed to `ConfigConstructable::new`, those typed fields are invisible to the constructor — hence workarounds like injecting a `provider_config` key into the blob at the callsite.

---

## Desired design: lazy memoised instance registry

Each factory (Provider, Model, Capability, Launcher) gains a memoised `construct_shared` method:

```rust
pub fn construct_shared(
    &self,
    name: &str,
    cfg: &serde_json::Value,
) -> Result<Arc<dyn Trait>, String>
```

The cache key is `(type_name, cfg_json_string)`. Two calls with identical arguments return the same `Arc` (pointer equality). This replaces construction-time dependency resolution with a lookup model: a `ModelSource` holds `provider_configs` at build time and resolves a live `Provider` on demand via `provider_for(model)` — no global config is baked into any struct.

`ConfigConstructable::new` drops to one argument:

```rust
fn new(cfg: &serde_json::Value) -> Self
```

Types that need cross-component access (e.g. `AgentModelCapability` needing its model's provider) receive the dependency at **use time** rather than **construction time** — e.g. `Capability::bind` accepts a `models: &dyn ModelConfigured` arg rather than storing an `Arc<dyn Model>` in the capability struct.

---

## What already exists

Branch `feat/Instance_Reference_Semantics` implements most of this:

- ✅ `construct_shared` with `Mutex`-based memoised `Arc` cache in `define_factory!`
[...]

PR #106, opened 2026-08-30
Plan: Config Reference Integrity & Remediation

afrittoli:
## Description

Plan: Config Reference Integrity & Remediation

## Changes

Add a plan file.

## Testing

Described in the design

## Related Issue(s)

#47, #90

## AI Usage

Initial draft based on requirements, diagram and UI design samples via Claude+Sonnet 5.

### DIFF
--- a/docs/specs/0024-config-integrity-validation.md
+++ b/docs/specs/0024-config-integrity-validation.md
@@ -1,0 +1,428 @@
+# Plan: Config Reference Integrity & Remediation
+
+## Overview
+
+Configured items reference each other by id (a capability's `model_id`, a
+model's `provider_id`, a launcher's `enabled_capabilities`), and nothing
+validates those references stay live. 
+
+Configuration may become inconsistent upon removal (see #47) which causes
+dangling references that may prevent from using one command until the issue
+is fixed (see #90). It can also become inconsistent during creation, when a
+nested "configure a new X" step silently fails to persist but is trusted
+anyway.
+
+This plan covers three aspects for the entire CLI:
+- how to detect inconsistencies in configuration
+- how to notify the user about them and help them fix such issues seamlessly
+- how to prevent new inconsistencies from being introduced while configuring
+  something else
+
+
+## Proposal
+
+We introduce a new `is_configured()` contract across `Model`, `Launcher`, and
+`Capability`, which verifies if a specific instance is consistently configured.
+
+- **`is_configured()`** checks are cascaded: must check the next hop's
+  `is_configured`, not just construction success.
+- **`is_configured()`** refers to static configuration only. Runtime checks
+  are performed by `health_check()` instead (e.g. is a provider running)
+
+```
+┌────────────────────────────┐
+│ Launcher::is_configured()  │
+└──────────────┬─────────────┘
+               │   every enabled capability must appear
+               │   in CapabilitySource(..).instances()
+               ▼
+┌────────────────────────────┐
+│ Capability::is_configured()│
+└──────────────┬─────────────┘
+               │   its model resolves, and that model
+               │   is itself is_configured()
+               ▼
+┌────────────────────────────┐
+│   Model::is_configured()   │
+└──────────────┬─────────────┘
+               │   provider_config() is Some
+               ▼
+┌────────────────────────────┐
+│      Provider exists?      │
+└────────────────────────────┘
+```
+
+`health_check()` (Sub-Task 7) walks the same chain but is a separate,
+I/O-bound check — never part of this cascade, only invoked by `launch` and
+`setup`.
+
+Nested "configure a new X" flows (e.g. picking a provider while setting up
+a model) can silently create a broken or unpersisted reference if the
+nested step fails to save. We make these flows re-validate what they
+created instead of trusting a bare `Ok(())`.
+
+We define a per-command policy for what to do about a broken reference
+once found (a list shouldn't interrupt with a prompt; a launch should).
+
+### Command policy
+
+| Command | On a broken reference |
+|---|---|
+| List (`model/capability/launcher/provider list`) | Inline annotation, never prompt |
+| Info/detail (`model/capability info`) | Warning + offer remediation |
+| Setup (create) | Remediation only for a dependency the wizard is about to use — see [open question](#open-questions)  |
+| Setup (overwrite of an existing instance) | Visually flag defaults that are themselves broken |
+| Launch | Offer remediation, abort by default if declined; then, separately, `health_check()` fail-fast (no remediation possible for a down service) |
+| Remove | Warn and offer a choice before creating a new dangling reference |
+
+Remediation is interactive-only, gated on **both** `Ui::is_interactive()`
+*and* the command not being in an auto/non-prompting mode (e.g. `setup
+--auto`) — `is_interactive()` alone only reflects the output backend, not a
+command-level flag. Warnings always show, routed to stderr for `json`/`markdown`.
+
+Note that a bug exists today, so `warn()` currently write to stdout for
+`json`/`markdown`. This needs to be fixed and sent to stderr instead.
+
+## Open Questions
+
+- Should a broken candidate appear at all in a `setup` selection list
+  (e.g. "Select a model for this capability")? Today `*Source::from_config`
+  filters it out of `instances()` entirely.
+- Should we implement multi-hop remediation within one session, when fixing
+  one reference reveals another.
+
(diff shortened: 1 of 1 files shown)


gabe-l-hart (commented):
A couple things jump out at me for further discussion:

1. We need the logic to be generic in `Capability` based on `dependencies()` since some capabilities will not need just a single model (some may need several, some none)
 
2. If I'm reading correctly, the `is_configured()` will be an instance method. This means that the every type that holds an instantiated instance of a dependent type will need to hold it as `Option<...>` to allow this unconfigured state to be acceptable. I'd really like to avoid this if possible and make sure the type is _configurable_ before instantiation. We already have logic for instantiating dependencies in the `commands`, but the error-path is to `panic` (or `anyhow::bail`). I think we could accomplish this pre-instantiation by improving the error handling when doing this kind of recursive instantiation for inputs to the instantiation of a given type. If gathering dependencies fails, we should be able to identify where it failed and use that as the signal to fix broken config.

afrittoli:
> A couple things jump out at me for further discussion:
> 
> 1. We need the logic to be generic in `Capability` based on `dependencies()` since some capabilities will not need just a single model (some may need several, some none)

That was the intention indeed, I'll fix the proposal to make that more explicit.
One possibility could be to go even further:
* Move `is_configured` to the dependency interface
* Change model and launcher to have dependencies, like capability does

> 2. If I'm reading correctly, the `is_configured()` will be an instance method. This means that the every type that holds an instantiated instance of a dependent type will need to hold it as `Option<...>` to allow this unconfigured state to be acceptable. I'd really like to avoid this if possible and make sure the type is _configurable_ before instantiation. We already have logic for instantiating dependencies in the `commands`, but the error-path is to `panic` (or `anyhow::bail`). I think we could accomplish this pre-instantiation by improving the error handling when doing this kind of recursive instantiation for inputs to the instantiation of a given type. If gathering dependencies fails, we should be able to identify where it failed and use that as the signal to fix broken config.

Good point, `is_configured()`  should not be an instance method. Two options:
* check during instantiation, as you suggested
* check before instantiation, through a cascading, static check

I'm still learning the code base, but I suspect either way this check could be hooked into `ConfigConstructable`.
Either case the caller needs to be able to identify the error and a "ConfigurationBroken" one, and handle it accordingly, depending on the command being executed.

docs/specs/0024-config-integrity-validation.md
```diff
@@ -0,0 +1,457 @@
...
+┌──────────────────┐
+│  LauncherConfig  │
+└──────────────────┘
+         │  enabled_capabilities: Vec<String>
+         │  wrapper field; no type knowledge needed
+         ▼
+┌──────────────────┐
+│ CapabilityConfig │
+└──────────────────┘
+         │  config["model_id"]
```
gabe-l-hart: This will still be impl specific. Anything in `config` should be considered opaque since this is where each impl can have its own different config schema. Instead, this should leverage the impl's metadata.dependencies. This is called out explicitly in the text below, so we just need to update the diagram.

docs/specs/0024-config-integrity-validation.md
```diff
@@ -0,0 +1,457 @@
...
+
+```
+capability setup chat     ->  chat, its model, that model's provider
+launch claude             ->  claude, its enabled capabilities,
+                              their models, those models' providers
+model list                ->  every model (the command's whole subject)
+provider remove ollama    ->  whatever currently points at ollama
+```
+
+The same check also runs inside construction, to remove the panic. The
```
gabe-l-hart, line 103: Nice, this is a good way to avoid the panic without needing to make everything optional

docs/specs/0024-config-integrity-validation.md
```diff
@@ -0,0 +1,457 @@
...
+
+The second is creation: a nested "configure a new X" step can report success
+without persisting anything, and an overwrite wizard can present a broken
+current value as a default that the user accepts without noticing.
+Both are handled by validating what was just produced rather than trusting
+it (Sub-Task 5).
+
+## Open Questions
+
+- Should a broken candidate appear at all in a `setup` selection list
```
gabe-l-hart: Good question. If we can detect that it's broken, I think we could present it with warning presentation (`ui.warn_mark`) and then force a reconfigure. This could also be a follow-on since it would be a bit of scope creep on this already large refactor.

docs/specs/0024-config-integrity-validation.md
```diff
@@ -0,0 +1,457 @@
...
+  (e.g. "Select a model for this capability")? Today `*Source::from_config`
+  filters it out of `instances()` entirely.
+- Should we implement multi-hop remediation within one session, when fixing
```
gabe-l-hart: I _think_ that would be best since without this, a user would need to run the offending command repeatedly until all remediation was done. If it turns out to be too hard to implement, we can punt, but I think if we remediate at the point of detection, we should be able to just recursively call the validation until it all passes.

docs/specs/0024-config-integrity-validation.md
```diff
@@ -0,0 +1,457 @@
...
+  `Capability` and `Launcher` a `health_check()` is its own piece of work.
+  Keeping it out is what lets validation stay pure and I/O-free. Note that
+  #36 proposes a status column at list time, the same surface Sub-Task 3
+  annotates, so the two want one column carrying both signals rather than
+  two columns.
+- Warning stream routing across the UI backends (#118). Every text backend
+  sends `warn()` to stdout and `error()` to stderr today, and the `Ui` trait
+  specifies stderr only for `error`. Whether `warn` should join it is a
+  policy question about the trait, not about config integrity.
+- `ModelSource` eagerness (#58). It builds every configured model on each
```
gabe-l-hart, line 172: Hmm, this is an interesting one. If we tackled this first, would the design of this solution fall back to construct-time remediation? If so, we should probably tackle them in order. There may be some benefit to decoupling dependency validation from construction, though, so in that case I think we can call this a design choice and move forward.

docs/specs/0024-config-integrity-validation.md
```diff
@@ -0,0 +1,457 @@
...
+  `ConfiguredModel::resolve()`, which is both wasted work and the reason
+  construction cannot be scoped. #58's lazy memoised `construct_shared` is
+  that fix; this plan works around it meanwhile by driving validation from
+  the command layer.
+- Malformed instance config (#59). `ConfigConstructable::new` deserialises
```
gabe-l-hart, line 177: Agreed that this is out of scope, but it's something we should certainly consider adding as a feature issue. We should see warnings if/when impl schema deserialization either misses fields or finds unused fields. We may also need to provide a mechanism for compatibility forward-mapping in the case that these config schemas ever break, but that's a problem for the future when it happens.

docs/specs/0024-config-integrity-validation.md
```diff
@@ -0,0 +1,457 @@
...
+  with `unwrap_or_default()`, so a corrupt config silently becomes a default
+  rather than an error. Validation here checks that references resolve, not
+  that each instance's own config parses.
+- Checking that a referenced model *satisfies* the `ModelRequirement` its
```
gabe-l-hart, line 181: Agreed, though this is also something I do want to tackle eventually.

gabe-l-hart (commented):
The design looks very solid now. One nit-pick on the initial diagram and one big-ish open question on whether we should tackle the `ModelSource` eagerness first and update the design to center on construction rather than pre-construction. My gut is to keep the validation and remediation independent, though. This would allow construction to essentially assert consistency and stay simpler that way.

docs/specs/0024-config-integrity-validation.md
afrittoli: Sounds good, I will update accordingly.

docs/specs/0024-config-integrity-validation.md
afrittoli: Agreed. Keeping validation and remediation decoupled from construction is a deliberate design choice, and we can lock that in and move forward.
I would then implement this spec first and then tackle issue #58.

afrittoli:
Thanks for the review @gabe-l-hart. I updated the plan accordingly (fixed the diagram, incorporated open questions which have been answered). I will create issues for the out-of-scope items and kick off the implementation.

gabe-l-hart (approved):
Looks great! @afrittoli I'm not sure if we want to merge this now as the plan of record and implement separately, or wait and include it with the implementation. Totally up to you.

PR #109, opened 2026-08-31
fix: Restore cursor on Ctrl-C during prompts

### LINKED ISSUE #93: bug: Terminal reset needed after Ctrl-C
## Description

When running through any of the configuration wizards, if the user uses `ctrl-c` (SIGINT) to interrupt the process while the user is using a `dialoger` menu, the terminal is left in a bad state and needs to be `reset`. We should figure out how to catch SIGINT and perform whatever reset is needed automatically before exiting.

aanokh:
## Description

Fixes #93 to restore cursor state properly on SIGINT

## Changes

Adds a SIGINT handler that restores the cursor and exits with code 130.

## Testing

Automatic test suite ran fully correct.
Test confirmed by hand to fix the ctrl C during setup bug.

## Related Issue(s)

#93 

## AI Usage

IBM Bob was used to help code this PR

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -448,6 +448,13 @@ impl alog::Formatter for UiFormatter {
 
 #[tokio::main]
 async fn main() {
+    tokio::spawn(async {
+        if tokio::signal::ctrl_c().await.is_ok() {
+            let _ = dialoguer::console::Term::stderr().show_cursor();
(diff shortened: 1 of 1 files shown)


gabe-l-hart (approved):
Nice! Works for me

PR #110, opened 2026-08-31
Feat: opencode full usage tracking

### LINKED ISSUE #86: feat: Full Usage Tracking in OpenCode
## Description

Currently, OpenCode only tracks usage for models that are managed by `granite-cli`. Unlike `claude`, there is no single URL override that can be used to capture traffic to all models, so a full implementation would need to parse all of the config surface for OpenCode (https://opencode.ai/docs/config/#precedence-order), then override all configured providers. It would also need to handle built-in providers for well known services (eg Anthropic, OpenAI).

## References

This was originally part of #82, but the scope got too big. I have a partial implementation saved at https://github.com/gabe-l-hart/granite-cli/tree/save/OpenCodeFullUsageTracking.

aanokh:
## Description

Adds usage tracking for OpenCode building on the WIP implementation from save/OpenCodeFullUsageTracking branch.
Since OpenCode doesn't have a single url override, the user configured providers bypassed usage tracking. This PR discovers user providers, and registers them to route through the proxy. 

Known limitations: remote/managed config layers and OAuth-authenticated built-in providers are not covered. Project-level opencode.json providers may override the rewrite due to OpenCode's config precedence.

## Changes

The PR is based off of the https://github.com/gabe-l-hart/granite-cli/tree/save/OpenCodeFullUsageTracking branch. In addition, provider routing was changed to be based on url path instead of model name prefix. This avoids situations where a gpt selected model in openrouter would be sent to openai api. Also fixed path formatting bug in the config discovery and added new tests.

## Testing

Full test suite passed, and ran manual tests with OpenCode and OpenRouter calls to gpt and gemini models.

## Related Issue(s)

Fixes #86 

## AI Usage

IBM Bob was used to implement the code in this PR.

### DIFF
--- a/src/launchers/opencode.rs
+++ b/src/launchers/opencode.rs
@@ -3,10 +3,12 @@
 //! Unlike `pi`, OpenCode's config sources are additive: `OPENCODE_CONFIG`
 //! points at one extra file that gets merged into the chain (loaded after the
 //! global config, before the project config) rather than a whole directory to
-//! redirect wholesale. So this launcher never touches the user's own
-//! `opencode.json`, credentials, or session store -- it only ever writes its
-//! own small generated file under `GRANITE_CLI_HOME` and points
-//! `OPENCODE_CONFIG` at it.
+//! redirect wholesale. This launcher reads the user's global and project
+//! `opencode.json` configs to discover configured providers, registers them on
+//! the session proxy for usage tracking, then writes its own small generated
+//! file under `GRANITE_CLI_HOME` and points `OPENCODE_CONFIG` at it. The
+//! generated config includes granite-cli providers plus any user providers
+//! merged in with their `baseURL` redirected to the proxy.
 
 // Standard
 use std::collections::HashSet;
@@ -26,6 +28,7 @@ use crate::capabilities::{
 use crate::launchers::base::{EnvBinding, LaunchContext, Launcher, LauncherMetadata, run_command};
 use crate::launchers::shared::mcp_cli::mcp_binding_request;
 use crate::providers::ApiType;
+use crate::proxy::ProxyHandle;
 use crate::registry::ConfigConstructable;
 use crate::utils::resolve_shell_command;
 use crate::utils::ui::Ui;
@@ -67,6 +70,11 @@ pub struct OpenCodeLauncher {
     /// its own `provider.<name>` entry and is referenced directly as
     /// `<provider>/<model>` in `agent.<name>.model` -- no mini-router needed.
     bound_sub_agents: Vec<(String, SubAgentBinding)>,
+    /// The session-scoped model proxy, if one was booted for this launch
+    /// (see `run_launch`) -- present whenever usage tracking or sub-agent
+    /// routing is needed. Used to redirect provider baseURLs so all traffic
+    /// flows through the proxy for usage accounting.
+    model_proxy: Option<ProxyHandle>,
 }
 
 impl ConfigConstructable for OpenCodeLauncher {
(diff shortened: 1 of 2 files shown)


src/launchers/opencode.rs
```diff
@@ -475,6 +522,471 @@ impl OpenCodeLauncher {
...
+            let Some(base_url) = resolved_base_url.as_ref().and_then(|b| b.as_str()) else {
+                continue;
+            };
+
+            // Skip if baseURL already points at the proxy
+            if base_url.contains("127.0.0.1") || base_url.contains("localhost") {
+                continue;
+            }
+
+            let label = format!("user-provider-{provider_name}");
```
gabe-l-hart, line 724: This is interesting. The `user-provider` prefix is pretty long and results in output that looks like this:

```sh
Binding                                              Requests  Input Tokens  Output Tokens  Cache Write  Cache Read
───────────────────────────────────────────────────  ────────  ────────────  ─────────────  ───────────  ──────────
granite-vision-4.1-4b                                1         0             0              0            0         
user-provider-pgx-direct/granite-exp-sft:Q4_K_M  10        97266         1632           0            84811     
Total                                                11        97266         1632           0            84811 
```

src/launchers/opencode.rs
```diff
@@ -475,6 +522,471 @@ impl OpenCodeLauncher {
...
+            }
+        }
+    }
+
+    /// Discovers env-based providers by checking for known API key environment
+    /// variables. Returns a map of provider name -> provider entry pointing at
+    /// the well-known base URL for each provider. These are providers that
+    /// OpenCode auto-loads when the corresponding env var is set, but don't
+    /// appear explicitly in the user's config files.
+    fn discover_env_providers() -> serde_json::Map<String, serde_json::Value> {
```
gabe-l-hart, line 743: This function feels like it could be a lot simpler if you extracted the repeated logic from each block into a helper that takes args for the `"npm"`, `"name"`, `"baseURL"`, and `"apiKey"` values.

gabe-l-hart: Also, I'm guessing this comes directly from OpenCode's source (eg [here](https://github.com/anomalyco/opencode/blob/51f86c853791c41656fb0adcf9413291e4996b87/packages/llm/script/setup-recording-env.ts#L170))? If so, we should cite it in comments so it's clear that this correlates 1:1 and would need to be updated if the set of supported providers changes.

Alternately, there's the `opencode providers list` command that we could possibly use to gather this information programmatically at runtime.

src/proxy/server.rs
```diff
@@ -255,8 +286,22 @@ impl RoutingTable {
...
-    fn target_and_label_for(&self, body: &[u8]) -> (ResolvedTarget, String) {
+    fn target_and_label_for(
+        &self,
+        path_and_query: &str,
+        body: &[u8],
+    ) -> anyhow::Result<(ResolvedTarget, String, String)> {
         let model = model_from_body(body);
+        if let Some((name, rest)) = split_provider_path(path_and_query) {
+            let Some((target, label)) = self.providers.get(name) else {
+                anyhow::bail!("no provider registered under '/providers/{name}'");
```
gabe-l-hart, line 297: I think this will break Claude which was relying on the defaulting behavior.

gabe-l-hart: Hm, it doesn't seem to. I guess Claude is still sending to the un-qualified provider path which does fall back to the default logic?

gabe-l-hart (changes requested):
This is a really nice improvement! I'd like to see if we can clean up some of the big code blocks, and I'd like to remove the `user-provider-{provider_name}` prefix since that just clutters up the output.

gabe-l-hart (approved):
Made the fixes and this is good to go now!

PR #113, opened 2026-09-01
Fix: Health checks now do proper JSON parsing

### LINKED ISSUE #92: bug: Bad health checks for llama.cpp and openai providers
## Description

The health checks for `llama.cpp` and `openai` simply check that they get ANY response rather than checking for a 2XX response. They both also use well-known ports, so if there's _any_ service running on those ports, they'll show up as healthy, even if it's something completely different. We should swap them to use `/v1/models` (a required part of the OpenAI REST API) and require a 200.

ByeolBaram:
## Description

Make proper health checks for llamacpp and openai.

## Changes

- Make the health check check for proper JSON payload and not just any response at all

## Testing

Manual Testing

## Related Issue(s)

Closes #92 

## AI Usage


gabe-l-hart (approved):
Confirmed that this fixes the false-positive healthy for both llama.cpp and generic OpenAI compatible server when there's another service running on 8080 that always returns a 200 (eg Open WebUI)

PR #114, opened 2026-09-01
fix: Detect WSL on windows

gabe-l-hart:
## Description

This PR detects if the install.sh script is running in WSL on Windows and downloads the windows binary instead of linux

## Testing

- [x] test on WSL

## AI Usage

built with pi and Qwen 3.6-35b


gabe-l-hart:
Tested on WSL and working for install and update

PR #117, opened 2026-09-01
feat: Add WSL/Windows fallback logic

gabe-l-hart:
## Description

There's a complex relationship between WSL and native-Windows. The compiled binaries in WSL can't run in native windows, but native windows binaries can run in WSL. How paths are parsed is compile-specific: if the binary is WSL- native, it uses unix paths, but if it's windows-native, even if run in WSL, it uses windows native paths.

The changes here are designed to enable the prebuilt windows binaries to run in both WSL and PowerShell. They cleanly enable running a windows-native OpenCode install (or other launcher) from either host shell. What they _don't_ enable (yet) is running a WSL-native launcher wrapped in granite-cli.exe from PowerShell. This is a reasonable limitation for now.

## Testing

- Built from WSL targeting native Windows

  ```sh
  cargo build --target x86_64-pc-windows-gnu --release
  ```

- Installed windows-native OpenCode through Chocolaty
- Configured `opencode` launcher to run with windows-native opencode.exe path
- Ran windows-native `granite-cli.exe launch opencode` from PowerShell (works as expected)
- Ran windows-native `granite-cli.exe launch opencode` from WSL (works as expected)
- Ran WSL-native `granite-cli launch opencode` from WSL (works as expected)

## AI Usage

This PR was built using OpenCode + Qwen3.6-35b on my Windows box. Commits were made with functionality review, but not code review. I will thoroughly review during PR.

### DIFF
--- a/src/config/mod.rs
+++ b/src/config/mod.rs
@@ -102,6 +102,17 @@ pub struct LauncherConfig {
 }
 
 impl Config {
+    /// Detect if running under Windows Subsystem for Linux by checking
+    /// `/proc/version` for WSL markers.
+    fn is_wsl() -> bool {
+        std::fs::read_to_string("/proc/version")
+            .map(|s| {
+                let lower = s.to_lowercase();
+                lower.contains("microsoft") || lower.contains("wsl")
+            })
+            .unwrap_or(false)
+    }
+
     fn config_dir() -> Result<PathBuf> {
         let val_res = std::env::var("GRANITE_CLI_HOME");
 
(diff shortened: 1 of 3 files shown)


gabe-l-hart:
One more! We need to be able to translate the command path between the two config surfaces now that they share a config space.

src/config/mod.rs
```diff
@@ -101,7 +101,159 @@ pub struct LauncherConfig {
     pub config: serde_json::Value,
 }
 
+/*-- Windows / WSL ----------------------------------------------------------*/
+
+/// Convert a Windows path string to a WSL path.
+///
+/// `C:\Users\gabel\AppData\Roaming\granite-cli\launchers` →
+/// `/mnt/c/Users/gabel/AppData/Roaming/granite-cli/launchers`
+#[allow(dead_code)]
```
gabe-l-hart, line 110: Needed for windows build path since this is not called there

gabe-l-hart:
Verified that this still works as expected on Mac

Issue #119, opened 2026-09-02
feat: Add `info` subcommands for everything besides `model`


gabe-l-hart:
## Description

Currently you can run `granite-cli model info <name>` to get the details of a configured model, but the other types don't have this subcommand. They should!

iamjaeholee:
Hi! I'd like to work on this.

gabe-l-hart:
Hi @iamjaeholee ! We'd welcome a contribution to add this. Thanks for digging into it.

afrittoli:
I was about to file this for launcher :)

gabe-l-hart:
Closed by #121

Issue #118, opened 2026-09-02
enhancement: output of warnings


afrittoli:
This is the current handling of output:

| backend | `warn()` | `error()` |
|---|---|---|
| terminal, TTY | stdout | stderr |
| terminal, non-TTY | stdout (delegates to plain) | stderr |
| plain | stdout | stderr |
| markdown | stdout | stderr |
| json | stream (stdout) | stream (stdout) |

In json, the warning is embedded in the json structure, so the resulting JSON is still parsable and the warning can be separated.
In markdown and terminal/plain, warnings go to stdout, so there's no way to separate them.

On the flip side, mixing warnings and errors in stderr means that warnings cannot be easily separated from errors, but I think that can be easily solved by logging the severity level along with the message (which already is the case).

I would recommend to:
- for terminal/plain: move warning to stderr. The user experience won't change.
- for markdown: use a markdown warning (see example below) or move to stderr

> [!WARNING]  
> Critical content demanding immediate user attention due to potential risks.

iamjaeholee:
Hi !, I'd like to work on this : -)

gabe-l-hart:
Closed by #122

Issue #36, opened 2026-08-04
feat: Validate launcher binary at launch and list time


Tanker2020:
## Background

Currently, launcher binary validation (`validate_command`) is only called explicitly at `setup` time. After setup the binary could be moved, deleted, or uninstalled and granite-cli would have no feedback until the `launch` command fails with an opaque OS error.

## Proposed behaviour

- **At `launch` time**: always call `validate_command` before spawning the subprocess. If validation fails, print a clear actionable error (e.g. `'claude' not found — re-run `granite-cli launcher setup claude` to update the path`).
- **At `list`/`info` time** (optional): show a ✓/✗ status column indicating whether the binary is currently reachable, so users can spot broken configs at a glance without running a launch.

## Context

The standalone `granite-cli launcher validate <id>` subcommand is being removed in PR #33 per reviewer feedback. Validation at `launch` time is the replacement for the "binary moved/deleted" scenario it was trying to solve.

Noted in review of PR #33 by @gabe-l-hart: _"Probably not for this PR, but we should have this for all types."_

PR #123, opened 2026-09-08
feat: config integrity validation (part 1)

afrittoli:
## Description

[Spec 0024](https://github.com/ibm-granite-community/granite-cli/blob/main/docs/specs/0024-config-integrity-validation.md) , Sub-Tasks 1 to 3: a scoped reference validator, a remediation prompt
built on the existing setup and remove commands, and the wiring that surfaces
both through `list`, `info` and `launch`. They ship together because nothing
calls the validator until the wiring lands.

## Changes

**`config::validation`** answers whether a configured instance's references
resolve, reading config and registry metadata and constructing nothing. It
checks ids and `*_type` names and recurses, so a missing provider four levels
below a launcher is reported rather than the first hop. Errors carry a typed
`Problem` and the instance holding the broken reference. Also drops
`Capability::dependencies(&self)`, leaving `metadata()` as the single place a
capability declares what it needs.

**`commands::remediation`** offers a fix for the first problem and re-validates
after each one. What removal means depends on how it was reached: a caller that
named the instance gets deletion, while remediation reached through a launcher
gets un-enabling, since a capability may be shared and `launch claude` should
not change more than the launcher it was asked about. A repair that changes
nothing drops out of the choices rather than ending the run. `Ui` gains
`is_interactive()`, so JSON and Markdown sessions never prompt.

**`list`, `info` and `launch` now act on validation.** The four list commands
flag every instance whose references do not resolve. `capability info` and
`model info` report the problem and offer a fix before rendering. `launch`
checks the launcher, the capabilities it enables, their models and those
models' providers before starting anything, offers a fix for the first problem
it finds, and aborts if you decline.

Two pre-existing failures this also fixes:

- Launcher setup and the unified `setup` wizard aborted the process with exit
  101 whenever any configured capability's model was missing. They now skip
  that capability and say why it is not listed.
- Reconfiguring a launcher silently dropped every broken capability from its
  enabled list. It now keeps them and names them.

`model list` now keeps a row for a model whose type is unknown, where it used to
drop it, replacing what `list_unknown_model_id_in_config_is_skipped` asserted.
`ModelCommands::info`, `CapabilityCommands::info` and `run_launch` become async.

Removal does not yet check its dependents, so deleting a capability leaves the
launcher enabling it with a dangling entry that the next `launch` offers to fix.
That check is Sub-Task 4, in the following PR.

## Testing

40 new test functions, 685 to 725; the suite goes from 705 to 745 passing. Line
coverage is 98.6% and 97.9% on the two new modules, 78.5% to 79.8% overall.
Validation does no I/O, so no test needs a network or a provider. fmt and clippy
are clean, and every commit compiles and passes the full suite on its own. Also
tested by hand against an isolated `GRANITE_CLI_HOME`: broken and missing
references, unknown types, the preflight and its exit codes, both non-interactive
backends, and catalog-id browsing.

## Related Issue(s)

To file once this merges:

- [ ] **Re-type a broken instance in place.** When an instance's `*_type` is not
      in the registry, typically after a catalog rename, the only repair offered
      is removal. Removing a model then breaks every capability pointing at it,
      cascading a rename into a rebuild. Changing the type while keeping the
      instance id would preserve every inbound reference, but it needs a
      per-kind type picker that does not exist as a reusable piece today.
- [ ] **The second confirmation when reconfiguring.** Choosing "Reconfigure
      'chat' now" leads into that instance's setup wizard, which immediately
      asks "Capability 'chat' is already configured. Overwrite? [y/N]" and
      defaults to no. That is two confirmations for one decision, and taking the
      default returns having changed nothing, leaving the reference broken.
      Fixing it means changing how the four setup commands treat an explicit
      instance id, which affects every caller of them.
- [ ] **Sibling problems a declined prompt hides.** Remediation reports one
      problem at a time and stops when the user declines, so a launcher with two
      broken capabilities only ever names the first. The list commands show all
      of them, but someone working through the prompt cannot see how many
      remain, or that a second exists at all. Reporting them needs gathering
      every error rather than stopping at the first, which is the collect-all
      walk Sub-Task 1 deliberately does not do.
- [ ] **Launch without a broken capability.** A launcher with several enabled
      capabilities cannot start while any one of them is broken: declining the
      preflight aborts, and there is no way to run with the rest. Simply
      acknowledging the problem is not enough, because the capability stays
      enabled and constructing it would hit #90. The repair is to drop the
      broken id from `enabled_capabilities` for that run only, which needs a
      fourth action on the preflight prompt.

Partly fixed:
- **#90**, the `ConfiguredModel::resolve` panic is not reachable anymore
  thanks to validation, but the panic is still in the code base. The full fix would
  be to make `ConfigConstructable::new` fallible.

Related, not addressed here:

- **#119**, `info` subcommands for everything besides `model`. `launcher` and
  `provider` still lack one, so while working through the preflight one problem
  at a time there is no way to see which capabilities a launcher enables. This
  PR makes that gap more noticeable without closing it.
- **#36**, runtime liveness. Whether a provider is actually reachable is a
  different axis from whether config references resolve, and keeping it out is
  what lets validation stay pure and I/O-free. #36 proposes a status column at
  list time, which is the same column this PR annotates, so the two should end
  up sharing one column rather than adding a second.
- **#58**, `ModelSource` eagerness. It builds every configured model on each
  `ConfiguredModel::resolve()`, which is wasted work that this PR does nothing
  about: validation is driven from the command layer and reads config directly,
  leaving the eager build in place.
- **#118**, warning stream routing. Every text backend sends `warn()` to stdout
  while the `Ui` trait specifies stderr only for `error()`. This PR adds
  warnings to `list`, `info` and `launch`, so more output now depends on that
  unresolved policy question.

## AI Usage

Drafted with Claude (Opus 5) and reviewed commit by commit; each commit carries
an `AI-usage: draft (Claude + Opus 5)` trailer. Design decisions and the manual
test pass were done by hand, and several behaviours above came out of that
testing.


gabe-l-hart:
Functionality is all looking good in my testing. Scenarios I tried:

1. Configure `pi` with `agent-model` -> `granite-4.2-8b` -> `ollama`
  a. Run fully configured
  b. Remove `granite-4.2-8b`
  c. Use `capability list` -> See warning
  d. Launch -> reconfigure `agent-model` by recreating `granite-4.2-8b` -> successful launch
  e. Remove `ollama`
  f. Use `model list` -> See warning
  d. Launch -> reconfigure `granite-4.2-8b` by recreating `ollama` -> successful launch

2. Configure `claude` with [`sub-agent-explore` -> `granite-4.2-8b` + `vision-mcp` -> `granite-vision-4.1-4b`] -> `ollama`
  a. Run fully configured
  b. Remove `ollama` (orphaning both models)
  c. Launch -> reconfigure `granite-4.2-8b` -> successful launch
    * **NOTE**: There's an interesting case here where both models became invalid, but once the first was fixed, the second became valid again, so I was never prompted to fix it. I think this is OK, but just a point of ambiguity on the best course of action.

src/capabilities/vision_mcp/mod.rs
```diff
@@ -126,19 +126,6 @@ impl Capability for VisionMCPCapability {
         "Exposes a vision-language model as an MCP server (compare/analyze images) for a launched coding agent to call."
     }
 
-    fn dependencies(&self) -> Vec<Dependency> {
```
gabe-l-hart, line 129: If I'm reading correctly, this (and elsewhere) is fully removing instance-level dependencies which were only different from type-level dependencies based on the `resolved_id`. Is that about right?

src/capabilities/base.rs
```diff
@@ -314,9 +313,13 @@ impl std::fmt::Display for CapabilityMetadata {
...
 /// A capability's declared dependency on a model, provider, or external shell
-/// command. `resolved_id` is `None` at the type level (catalog display,
-/// before any instance is configured) and `Some(id)` once a concrete
-/// instance has picked a specific dependency.
+/// command, as declared by `CapabilityMetadata` for a capability type.
+/// `config_key` names the key in a configured instance's own config JSON that
+/// holds the resolved id, which is how `config::validation` reads it without
+/// constructing anything.
+///
+/// `resolved_id` has no producer since the instance method that filled it was
```
gabe-l-hart: This comment seems a bit AI-ish with a reference to something that was removed.

src/commands/launcher.rs
```diff
@@ -232,6 +239,33 @@ impl LauncherCommands {
...
 
+    /// Configuration integrity check for `launch`: validates the launcher,
+    /// the capabilities it enables, and what those resolve to, offering a fix
+    /// for anything broken.
+    ///
+    /// Declining aborts the launch rather than skipping, because a capability
+    /// that cannot bind would fail later during the launch itself. This runs
+    /// before anything about the environment, so a configuration problem is
+    /// reported before a missing binary is.
+    pub async fn preflight(ctx: &mut crate::AppContext, launcher_id: &str) -> Result<()> {
```
gabe-l-hart: NIT: I might prefer the name `prelaunch` since we don't use the "fly" metaphor, but I suppose rockets fly once they launch, so very weakly held.

src/commands/mod.rs
```diff
@@ -3,6 +3,7 @@ pub mod hardware;
 pub mod launcher;
 pub mod model;
 pub mod provider;
+pub mod remediation;
```
gabe-l-hart: I wonder if we should stick this in `utils` or `config`? Prior to this, all mods within `commands` correlate 1:1 with top-level sub-commands in the CLI.

gabe-l-hart:
One NOTE during testing/review: I see that the `dangling_notes` are not represented in the interactive TUI currently.

src/registry/mod.rs
```diff
@@ -202,6 +202,18 @@ macro_rules! define_factory {
 
                 /// Construct an instance by name with the given configuration.
                 ///
+                /// Two kinds of caller reach this, and it cannot tell them apart.
```
gabe-l-hart: This is an interesting informative comment, but I'm not sure it belongs here. This seems like it probably belongs in a spec somewhere instead. This module tries its best to not know anything about how it's used downstream (aside from being implemented to meet the needs of the downstream users).

src/main.rs
```diff
@@ -705,7 +705,13 @@ async fn run_launch(
     use crate::proxy::ProxyServer;
 
     // Load config fresh so we always pick up the latest saved state.
-    let mut config = crate::config::Config::new()?;
+    ctx.config = crate::config::Config::new()?;
+
+    // Configuration integrity first, before anything about the environment.
+    LauncherCommands::preflight(ctx, launcher_id).await?;
```
gabe-l-hart: Would change `preflight` -> `prelaunch` if we do that rename

src/config/validation.rs
```diff
@@ -0,0 +1,826 @@
...
+//! name that resolves to nothing is the same kind of inconsistency either
+//! way.
+//!
+//! One walk covers every kind. Each of the four config types implements
+//! [`Validatable`] to say what its type name is and which ids it points at,
+//! and [`validate`] does the rest.
+
+use std::collections::HashMap;
+
+use crate::capabilities::Dependency;
```
gabe-l-hart, line 15: I think we're doing some dependency inversion here. `config` should be the bottom of the heap for internal dependencies (depends on nothing else internal). This makes me think that all of this should move into `dependency` which is a module where it would be appropriate to depend on `config` as well as the other known-type modules (`capabilities`, etc)

src/config/validation.rs
```diff
@@ -0,0 +1,826 @@
...
+use crate::config::{
+    CapabilityConfig, Config, ConfigId, LauncherConfig, ModelConfig, ProviderConfig,
+};
+
+/*-- public --------------------------------------------------------------------*/
+
+/// The four kinds of configured instance that reference each other.
+#[derive(Debug, Clone, Copy, PartialEq, Eq, Hash)]
+pub(crate) enum RefKind {
```
gabe-l-hart, line 24: Hm. It feels like this list _should_ exist somewhere already, but maybe it doesn't? We have `crate::capability::base::Dependency` which has _some_ of this, but only from the perspective of a `Capability`, so `Launcher` and `Capability` itself are not present.

src/config/validation.rs
```diff
@@ -0,0 +1,826 @@
...
+
+/// What went wrong with a reference. Callers branch on this to decide what to
+/// offer the user, rather than matching on a rendered message.
+#[derive(Debug, Clone, PartialEq, Eq)]
+pub(crate) enum Problem {
+    /// The named instance is not in the configuration.
+    NotConfigured,
+    /// The instance's `*_type` is not a key in its kind's registry.
+    UnknownType { type_name: String },
+    /// A model that names no provider at all, as distinct from one naming a
```
gabe-l-hart: When is this possible? Seems like this should be programmatically impossible? I see that `ModelConfig.provider_id` is `Option<String>`, but that seems like a problem. Can we explore making provider fully required for models?

gabe-l-hart: I went down this rabbit hole and I think it's doable to make this fully non-optional. I have local changes that I'll add to this branch.

src/config/validation.rs
```diff
@@ -0,0 +1,826 @@
...
+/// through whatever it finds. A launcher walks its enabled capabilities,
+/// their models, and those models' providers, so a missing provider four
+/// levels down is reported rather than the walk stopping at the first level.
+///
+/// The walk covers only what it was asked about. Nothing here reads a part of
+/// the configuration the caller did not name.
+///
+/// # Examples
+///
+/// ```ignore
```
gabe-l-hart: This seems a little brittle to have this code block in comments

src/config/validation.rs
```diff
@@ -0,0 +1,826 @@
...
+
+    /// The instances this one points at, for the walk to follow.
+    ///
+    /// The error is for an instance that cannot name its references at all,
+    /// such as a model with no provider. That is a problem with this
+    /// instance rather than with anything it points at.
+    fn refs(&self) -> Result<Vec<(RefKind, &str)>, Problem>;
+}
+
+/// Reaching the four maps by [`RefKind`] rather than by name. This lives with
```
gabe-l-hart: Interesting. Having this here instead of in `mod.rs` was pretty confusing. I think we might want to pull some of this over there or write these as free-functions that take a `Config` argument.

src/commands/remediation.rs
```diff
@@ -0,0 +1,804 @@
...
+    /// The instance's `*_type`, which reconfiguring hands back to setup.
+    type_name: String,
+    /// False when the type name is itself the problem. Setup cannot run a
+    /// type the registry does not have, so removal is the only fix.
+    can_reconfigure: bool,
+    /// The `(launcher, capability)` pair to un-enable, when remediation was
+    /// reached through a launcher that enables the capability. Some means the
+    /// removal on offer drops the id from that launcher's list rather than
+    /// deleting the instance.
+    unenable: Option<(String, String)>,
```
gabe-l-hart: `disable`?

src/commands/remediation.rs
```diff
@@ -0,0 +1,804 @@
...
+/// A capability is shared: other launchers may enable the same instance, and
+/// the caller asked to launch one launcher rather than to change the
+/// configuration at large. Dropping the id from that launcher's own list
+/// repairs what was asked about and leaves everything else alone.
+///
+/// Two shapes reach here. The launcher enables a capability whose own
+/// reference is broken, where the fix acts on that capability; and the
+/// launcher enables a capability that is not configured at all, where the fix
+/// acts on the launcher.
+fn unenable_target(
```
gabe-l-hart: `disable_target`?

gabe-l-hart (changes requested):
This is a big overhaul, but it's great! I've got a few naming/code-organization thoughts, and I'm going to push a change to make `ModelConfig.provider_id` not optional, but functionality-wise, I think it's good-to-go, so we should be ready once the organizational questions are sorted out.

afrittoli:
> One NOTE during testing/review: I see that the `dangling_notes` are not represented in the interactive TUI currently.

Good catch - I did not do anything on the TUI - happy to fix that as a follow up if that's ok.

src/capabilities/vision_mcp/mod.rs
afrittoli: Yes, I dropped instance level deps as - at least for now - they did not provide anything in addition to what the static version already provided. If there is a use case for having the distinction in future we can bring them back.

src/commands/launcher.rs
afrittoli: happy to switch to `prelaunch`

src/capabilities/base.rs
afrittoli: sorry I missed that - will fix

src/commands/mod.rs
afrittoli: Yes, I was debating the same. 
I'm not a big fan of utils modules as they risk to end-up a mix of a lot of different things. 
What about a having `commands.utils` folder and the remediation module within?

src/registry/mod.rs
afrittoli: I think it may be helpful here from someone calling this API to understand how to use it, but happy to add it to a development.md or api.md or similar if you prefer.

src/config/validation.rs
afrittoli: That'd be great, it would definitely simplify things.

src/config/validation.rs
afrittoli: Yes, the problem I had was that dependencies is not used everywhere in the data model. Certain things are connected via dependencies, others aren't. 
One of the options I considered was to migrate everything to using dependencies, but that was a bigger change and scope creep on this spec, so I decided against it for spec 0024. But that's a refactor we can still do if you want. 

I asked Claude about it, and with the size of the code and data model we have today, it would not really reduce the amount of code, but it might make it more linear to understand.

src/config/validation.rs
afrittoli: See my comment above https://github.com/ibm-granite-community/granite-cli/pull/123#discussion_r3972118338

src/config/validation.rs
afrittoli: Do you mean that if the code change the example there would become stale?
I do know if rust has any way to keep tested examples in docs or code, I can investigate.

src/commands/remediation.rs
afrittoli: Err... yes!

src/config/validation.rs
afrittoli: ok, sounds good.

Claude proposed having dedicated methods `validate_launcher`, `validate_model` etc, but I wanted to have instead the `Validatable` trait and a single `validate_ref` for all, so `entry` and `entries` got created in situ to help with that, but it makes sense to move them over.

Is there anything else you'd like to see move over?

src/capabilities/vision_mcp/mod.rs
gabe-l-hart: Agreed, let's get rid of it for now

src/commands/mod.rs
gabe-l-hart: Yeah, I agree. I always start with a `utils` which ends up having a super nasty mental dependency tree and only the core authors know what gets used where. I think a `commands.utils` is a good idea since remediation is fundamentally scoped to `commands`.

src/registry/mod.rs
gabe-l-hart: I agree there's good information here. Maybe we can just make this a bit more generic and not reference other modules in the codebase directly? Maybe just describe the two usage patterns as construction from a saved config object versus ephemeral construction for inspection? I never liked that double-construct pattern in the first place.

src/config/validation.rs
gabe-l-hart: Yeah, I'm debating the same thing. The `dependencies` stuff is used at setup-time to resolve unsatisfied dependencies that should be filled in while setting up a downstream type (eg `setup capability agent-model` -> `model` -> `provider`). This logic is _close_ to what you've got in this refactor, and ideally I'd love to consolidate the two, but I agree that it's probably best to keep this refactor more narrowly scoped for now.

At the moment, `capabilities` are the only trait group that have undefined dependency sets and therefore are the only ones that store their dependency pointers in the opaque `config` json. My intent had always been (like the factory/registry) to have a single abstract implementation of the dependency resolution framework that did type inspection for anything marked as a `Dependency`. It's conceivable that certain "Model" types could actually depend on multiple providers (eg a routed virtual model), so the explicit dependency keys for `Model` and `Launcher` were never really intended to be quite as static as they've become.

src/config/validation.rs
gabe-l-hart: Yeah, that's what I mean. I've seen this pattern of inline commented examples pretty often with Claude, but it always feels brittle to me. No strong preference, though. The one real reason I'd probably prefer removing the `# Examples` blocks is that we don't have them consistently across public interfaces and it would create a bit of style confusion for contributors or agents.

src/config/validation.rs
gabe-l-hart: As I was writing that comment, I went back-and-forth. I actually think putting them in `mod.rs` risks some leakage since their signatures will require `Validatable` and `RefKind`. I think it would work to just do `fn config_entry(config: &Config, ...)` and have these two functions as free-functions here rather than members on `Config`.

src/commands/mod.rs
```diff
@@ -4,6 +4,7 @@ pub mod launcher;
 pub mod model;
 pub mod provider;
 pub mod setup;
+pub mod utils;
```
gabe-l-hart: I should have caught this earlier, but we have `launchers/shared`, so I think we should maybe stick with that naming convention for this (`commands/shared`)?

gabe-l-hart (changes requested):
One quick rename request again (sorry I missed this earlier)

src/commands/mod.rs
afrittoli: Done

src/config/validation.rs
gabe-l-hart: For now, I think we can avoid over-fixing here, but we should open an issue to clearly define the relationships between modules in the project, and this should at a minimum include moving `capabilities::requirements` to something more generic. I would say we should put it in `dependency`, but this too will create a cycle. For now, let's just add a TODO comment and move on:


```suggestion
// TODO: This is a circular dependency that needs to be untangled
use crate::capabilities::Dependency;
```

gabe-l-hart (approved):
One more suggestion. To close the last comment without blowing up the scope of the PR, let's just note the circular dependency with a TODO. Otherwise, this is good to go!

PR #128, opened 2026-09-09
feat: config integrity validation (part 2)

afrittoli:
## Description

[Spec 0024](https://github.com/ibm-granite-community/granite-cli/blob/main/docs/specs/0024-config-integrity-validation.md),
Sub-Tasks 4 and 5: a removal now checks what it would strand, and a setup step
can no longer leave a reference pointing at nothing. Stacked on #123.

The top 3 commits belong to this PR. The rest is PR #123.
I will rebase once #123 is merged.

## Changes

**Removal asks about its dependents.** The four removal commands scan for the
instances pointing at the target and offer to remove them together, cancel, or
remove only what was asked. Cancel is the default, since the other two both
delete something. A session with nobody to ask removes only what was asked and
warns. Dependents go through their own removal commands, so the question
repeats down the reference graph. Finding them is `Validatable::refs` read
backwards, next to the walk in `config::validation`.

**Setup checks what it actually configured.** `Config::insert_*` updates the
in-memory map before saving and its callers treat a save failure as a warning,
so a nested setup step could report success over configuration that was never
written. `resolve_model_dependency` now keeps only the ids setup left behind
that validate. `select_provider`, which checked nothing, now rejects an id
configured as a different provider type, which is what a declined overwrite
onto a name collision used to return.

**Selection prompts name the instance currently configured** and say when it
no longer resolves, so pressing Enter through an overwrite cannot re-save a
dangling reference.

Also adds the spec's implementation summary, covering all five sub-tasks.

## Testing

11 new test functions, 740 to 751; the suite goes from 760 to 771 passing.
Line coverage is 98.7% on `config/validation.rs` and 97.9% on
`commands/remediation.rs`, 80.1% to 80.4% overall. fmt is clean, clippy adds
nothing new, and both feature commits pass the full suite on their own.

## Related Issue(s)

To file once this merges:

- [ ] **Removing a capability without breaking or deleting its launcher.** The
      prompt's three answers delete the launcher too, do nothing, or leave it
      pointing at something gone. A fourth would drop the id from
      `enabled_capabilities` as part of the removal; the behaviour exists in
      `remediation::unenable` and needs extracting.
- [ ] **A flag for non-interactive removal.** A script cannot ask for the
      dependents to go too, or for the removal to be refused with a non-zero
      exit when anything depends on the target.

## AI Usage

Drafted with Claude (Opus 5) and reviewed commit by commit; each commit
carries an `AI-usage: draft (Claude + Opus 5)` trailer.

### DIFF
--- a/src/config/validation.rs
+++ b/src/config/validation.rs
@@ -109,6 +109,47 @@ pub(crate) fn find_dangling(kind: RefKind, config: &Config) -> Vec<DanglingRef> 
         .collect()
 }
 
+/// The configured instances that point at `(kind, id)`, which is what
+/// removing it would strand. Sorted, so a caller listing them is stable.
+///
+/// This is [`Validatable::refs`] read backwards: an instance depends on the
+/// target when the target appears among the references it declares.
+///
+/// # Examples
+///
+/// ```ignore
+/// // Before `model remove granite-3.1-8b-instruct` deletes anything.
+/// let stranded = dependents(RefKind::Model, "granite-3.1-8b-instruct", &config);
+/// // -> [(RefKind::Capability, "chat")]
+/// ```
+pub(crate) fn dependents(kind: RefKind, id: &str, config: &Config) -> Vec<(RefKind, String)> {
+    let mut found: Vec<(RefKind, String)> = [
+        RefKind::Launcher,
+        RefKind::Capability,
+        RefKind::Model,
+        RefKind::Provider,
+    ]
+    .into_iter()
+    .flat_map(|referrer_kind| {
+        config_entries(config, referrer_kind)
+            .into_iter()
+            .map(move |entry| (referrer_kind, entry))
+    })
+    .filter(|(_, entry)| {
+        // An instance that cannot name its references, such as a capability
+        // missing a required dependency, points at nothing.
+        entry.refs().is_ok_and(|refs| {
+            refs.iter()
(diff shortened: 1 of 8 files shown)


src/config/validation.rs
```diff
@@ -109,6 +109,47 @@ pub(crate) fn find_dangling(kind: RefKind, config: &Config) -> Vec<DanglingRef>
...
+/// # Examples
+///
+/// ```ignore
+/// // Before `model remove granite-3.1-8b-instruct` deletes anything.
+/// let stranded = dependents(RefKind::Model, "granite-3.1-8b-instruct", &config);
+/// // -> [(RefKind::Capability, "chat")]
+/// ```
+pub(crate) fn dependents(kind: RefKind, id: &str, config: &Config) -> Vec<(RefKind, String)> {
+    let mut found: Vec<(RefKind, String)> = [
+        RefKind::Launcher,
```
gabe-l-hart, line 127: I wish there were a cleaner way to do "all entries in the enum".

gabe-l-hart: Looks like the recommended implementation of this is `strum`: https://docs.rs/strum/latest/strum/

gabe-l-hart (approved):
Code looks good. Works as advertised. Ship it!

src/config/validation.rs
gabe-l-hart: Probably not worth adding the dependency just for this though.

PR #137, opened 2026-09-14
Recommended Configurations

gabe-l-hart:
## Description

This PR adds support for `RecommendedConfiguration`s and add pre-built recommendations. The goal is to give users an out-of-the-box config experience that will yield good results on their available system.

## Changes

- Add `resources/recommended_configs` for static yaml configs
- Add build-time codegen for static recommended configs
- Add `src/config/recommended_config.rs` module holding logic for analyzing the set of recommended configs (static and user-provided) against the available system hardware and installed tools.
- Overhaul `src/commands/setup.rs` to use recommended configs in both auto and wizard modes

## Testing

- Unit tests
- Manual testing for auto mode:

  ```sh
  rm -rf tmp/*
  GRANITE_CLI_HOME=$PWD/tmp cargo run -q setup --auto
  ```

  - `claude`: `sub-agent-explore` w/ `granite-4.2-3b` + `vision-mcp` w/ `granite-vision-4.1-4b`
  - `opencode`: All of `claude` + `agent-model` w/ `granite-4.2-8b` (30b doesn't fit with min context)
  - all others: `agent-model` w/ `granite-4.2-8b` and/or `vision-mcp` w/ `granite-vision-4.1-4b`

- Manual testing for wizard mode:

  ```sh
  rm -rf tmp/*
  GRANITE_CLI_HOME=$PWD/tmp cargo run -q setup
  ```

  - Select `claude`, `opencode`, `goose` agents
  - Only pre-select `agent-model`, `sub-agent-explore` and `vision-mcp`
  - Propose models `granite-4.2-8b`, `granite-4.2-3b`, `granite-vision-4.1-4b`
  - Select variants and local Ollama runtime
  - Verify correct models map to correct capabilities map to correct launcers

## Related Issue(s)

Closes #129 

## AI Usage

The majority of this was developed using Claude Code launched through `granite-cli` with `sub-agent-explore` using `granite4.2:3b` in Ollama, and both `sub-agent-plan` and `sub-agent-code` using Qwen3.6-35b running in `llama-server`. Unfortunately, I hit ctrl-c too aggressively and lost the usage stats, but it very effectively used both `sub-agent-explore` and `sub-agent-code` to delegate work. I still haven't seen CC choose to use `sub-agent-plan`.

```sh
╔══════════════════════════════════════════════════════════╗
║           GIT AI USAGE ANALYSIS                          ║
╚══════════════════════════════════════════════════════════╝

📊 COMMITS BY AGENT

--- Aggregate ---
[...]

PR #140, opened 2026-09-16
Follow-up fixes to #137

### LINKED ISSUE #129: feat: Recommended Configurations
## Description

The current `granite-cli setup --auto` mechanism is overly aggressive in enabling all `Capabilities` for all `Launchers` as long as there's a model that can satisfy the launcher's dependencies and a provider to run it. This results in automatic configuration that provides bad results (eg running `claude` with `granite4.2:3b` in Ollama as `agent-model`).

The goal of this feature request is to implement a system where we can support well-known configuration parameters to constrain automatic setup to working setups. These well-known configurations should be based on a portable schema so that we can create static configurations in `resources` that get compiled into the system as well as allowing users to create and share their own that get loaded at runtime. A given configuration should be able to specify the following:

1. The `launcher` it applies to
2. The set of `capabilities` that _can_ be enabled if their recommended models are available
3. The set/range of `models` that can support each `capability`. This should allow model constraints to be specified based on any fields in `ModelMetadata`, but should focus on `family`, `version`, `size`, `context_length`, `supported_functions` (eg for optional functions like `Thinking`), and the availability of `varients` at specific precisions (eg "nothing lower than Q4 for quantization")

afrittoli:
## Description

Five follow-up fixes to #137, in the setup wizard and `setup --auto`. Each one
is a case where the recommended configurations either bound a capability to the
wrong model, configured a capability nothing could use, or overwrote an existing
configuration.

## Changes

- fix: Use recommended capabilities for the wizard ([e79919c](https://github.com/ibm-granite-community/granite-cli/commit/e79919c)) — each capability model slot takes the first model its launcher's recommended config lists among the models the user kept, and the fallback now reads the models in sorted order.
- fix: Skip capabilities a launcher cannot bind in setup --auto ([1832f0b](https://github.com/ibm-granite-community/granite-cli/commit/1832f0b)) — limits the capabilities `setup --auto` configures to those a detected launcher can bind, and with them the models those capabilities name.
- fix: Keep configured providers, models and capabilities in setup --auto ([9de25e7](https://github.com/ibm-granite-community/granite-cli/commit/9de25e7)) — resolves against configured providers as well as the healthy ones discovery found, and leaves an already configured model or capability as it is.
- fix: Recommend nothing for a launcher whose user config does not load ([244502f](https://github.com/ibm-granite-community/granite-cli/commit/244502f)) — a recommended config file that fails to parse no longer falls back to the built-in entry it replaces, unknown fields are rejected, and every skipped config file is logged with its parse error.
- fix: Skip a capability when any of its required model slots is empty ([26b61dc](https://github.com/ibm-granite-community/granite-cli/commit/26b61dc)) — one warning, and no entry written with a slot missing.

## Testing

Unit tests in each commit, covering model binding per capability, the binding
type gate for `bob` and `pi`, resolution through a configured provider, the
unknown-field rejection, and a capability with an empty required slot.

`scripts/manual-129/` holds a manual plan for the five fixes: fixtures that
write a scratch config, stub launcher binaries and a fake Ollama, then five
scenarios run through `setup` and `setup --auto`. 

## Related Issue(s)

- Fixes #129
- Follows up #137

## AI Usage

Drafted with Claude (Opus 5): the fixes, their tests and the commit messages
were drafted in a review session, then read, corrected and amended by hand.


src/commands/setup.rs
```diff
@@ -1251,25 +1337,69 @@ impl SetupCommands {
...
+            &selected_launchers,
+            &selection.providers,
+            &selection.models,
+            &selection.variants,
+            &selection.resolved_capability_models,
+            &selection.recommended_capability_types_by_launcher,
+        )
+        .await?;
+
+        // Never auto-pull in --auto mode
```
gabe-l-hart, line 1359: We might want to consider adding a `--pull` flag to `setup` that would allow `--auto` to pull as well (and skip the "do you want to" check for non-auto). Without pulling, `auto` still leaves the user unable to actually run anything.

gabe-l-hart (approved):
One thought for a follow-up around a `--pull` flag. Also, it seems from the PR body that there was a manual testing script created (`scripts/manual-129/`), but never pushed (which is good). Can you attach it as a comment to this issue?

afrittoli:
> One thought for a follow-up around a `--pull` flag. Also, it seems from the PR body that there was a manual testing script created (`scripts/manual-129/`), but never pushed (which is good). Can you attach it as a comment to this issue?

The test plan and fixtures are available in this commit:
https://github.com/afrittoli/granite-cli/commit/fb94828a4f1fd13e93da59a29c62768863f8421d

PR #142, opened 2026-09-16
feat: add --pull flag to setup

### LINKED ISSUE #141: feat: Add `--pull` to `setup`
## Description

https://github.com/ibm-granite-community/granite-cli/pull/140#discussion_r4028985192

The goal is to add `--pull` to `granite-cli setup`. When specified with `--auto`, it will automatically pull models. When specified without `--auto`, it will skip the user prompt and automatically pull models.

Voyagerroc-Lab:
## Summary
- add --pull to granite-cli setup
- pull selected models without prompting in wizard mode
- allow --auto --pull to pull configured models non-interactively
- keep --skip-pull available for explicitly skipping the pull phase

Closes #141

## Validation
- git diff --check
- Rust tests could not be run because Cargo is not installed in the execution environment.


src/commands/setup.rs
```diff
@@ -1178,16 +1178,21 @@ pub struct SetupCommands;
 
 impl SetupCommands {
     /// Entry point for `granite-cli setup`.
-    pub async fn run(ctx: &mut crate::AppContext, auto: bool, skip_pull: bool) -> Result<()> {
+    pub async fn run(
+        ctx: &mut crate::AppContext,
+        auto: bool,
+        pull: bool,
```
gabe-l-hart, line 1184: I think we should consolidate these two into a single `pull: Option<bool>` that gets set based on `--pull` / `--skip-pull` (which should be mutually exclusive). If `Some`, it should automatically take the specified behavior. If `none`, it should prompt in interactive mode and not pull in auto mode.

src/main.rs
```diff
@@ -535,15 +539,19 @@ async fn main() {
                 .await
                 .map_err(|e| ctx.ui.error(&e.to_string()))
         }
-        Some(Commands::Setup { auto, skip_pull }) => {
+        Some(Commands::Setup {
+            auto,
+            pull,
```
gabe-l-hart, line 544: I think we should consolidate `pull` and `skip_pull` into a single logical flag. The two would be mutually exclusive (I'm not sure if there's a way to do this in the `clap` framework) and would be optional (`none` if neither set).

gabe-l-hart (changes requested):
Thank you for the contribution! I think there's a nice simplification we should add to make `--pull` and `--skip-pull` aware of each other to avoid user confusion.

Also, please add an `## AI Usage` statement in the PR body to clarify what (if any) AI was used when creating this PR. We do accept agent-created PRs provided that the author can own the contribution and explain the use of AI during the process.

gabe-l-hart:
You should be able to get `fmt` clean with `./scripts/fmt.sh` provided you have a rust dev environment (the PR body suggests you don't). If you need a container image for a sandbox, see https://github.com/ibm-granite-community/granite-cli/blob/main/scripts/launcher-testing/Dockerfile

PR #146, opened 2026-09-16
fix: fail setup when the config cannot be saved

### LINKED ISSUE #125: Setup commands report success and exit 0 after failing to save
All four `*Commands::setup` functions treat a failed `insert_*` as a warning and then "print '<id>' configured successfully!" unconditionally, exiting 0. With an unwritable config directory the run prints both "failed to save provider config: ..." and "configured successfully!", and nothing is on disk.

Propagating the error instead would end the command with a non-zero exit and no success line, and would abort the outer setup rather than creating the dangling reference. Note this changes launch too, since remediation reaches setup through reconfigure.

maskjelly:
Closes #125.

All four `*Commands::setup` functions treated a failed `insert_*` as a warning and then printed `"<id> configured successfully!"` unconditionally, exiting 0 over configuration that never reached disk. With an unwritable config directory the run printed both `failed to save provider config: …` and `configured successfully!` — and callers that then referenced the new entry were left holding a dangling reference.

Each insert now propagates its original message with `map_err(…)?`, so the success line is unreachable when the save fails:

- `src/commands/provider.rs` — `insert_provider`, before the health check and success line
- `src/commands/launcher.rs` — `insert_launcher`
- `src/commands/model.rs` — `insert_model`
- `src/commands/capability.rs` — `insert_capability`

**Callers, verified for the issue's "would abort the outer setup" claim:**

- CLI: `main.rs` maps every command arm through `.map_err(|e| ctx.ui.error(&e.to_string()))` and exits 1 (`main.rs:488-591`), so the run now ends non-zero with a visible error.
- TUI: `utils/ui/app.rs:1522-1533` renders `Setup failed: {e}` in the pane (the pane stays alive, which is right for a TUI).
- Remediation: `shared/remediation.rs:385-391` is a pure pass-through and its caller propagates with `?` (`:231`); the upstream `launch`/`info` paths propagate too, so `launch` aborts instead of creating the dangling reference.

**Tests.** Each of the four setups gets a `#[cfg(unix)]` regression test that makes the `TestConfigHome` tree read-only, calls setup, and asserts both `result.is_err()` and that no captured info contains "configured successfully". Fail-without-fix (source reverted, tests kept): all four fail on `assertion failed: result.is_err()`. With the fix: 4/4 pass; full suite `880 passed, 0 failed`.

**Two things to look at, deliberately disclosed:**

1. **The test harness gained `Config::ensure_directories()`** (`src/config/mod.rs`, `TestConfigHome::new`). Without it, five *pre-existing* tests (`setup_agent_model_persists_config`, `info_reconfigures_a_broken_capability_when_asked`, `setup_custom_model_with_no_variants_…`, and two remediation tests) fail once the swallowed errors become real: the temp home has no `models/`, `providers/`, … directories, so their saves fail — they had been passing only because the bug hid it. I verified both directions (harness change disabled → those tests fail; restored → pass). If you would rather the harness stay minimal, the alternative is creating the needed directories inside each affected test.
2. **`make_unwritable()`/`make_writable()` are recursive.** Chmod'ing only the home root does not block writes through a still-writable `home/models/`, so the helper locks the whole tree and tests restore permissions before the `TempDir` guard drops.

**Not done, deliberately.** The bulk-wizard warnings in `src/commands/setup.rs` (`:2220`, `:2251`, `:2312`, `:2363`) are left as they are — the issue scopes the four `*Commands::setup` functions; say the word if you want the wizard to fail the same way.

**Gates** (with CI's pinned 1.88.0): `cargo fmt --all -- --check`, `cargo clippy --all-targets --all-features -- -D warnings`, `cargo test --locked --all-targets --all-features` — all clean.


gabe-l-hart:
The code looks good, but I'm also seeing this unit test failing locally. The test seems like it should be unrelated (it's related to the interactive TUI's show/hide functionality that was changed in #148 ). I'll investigate further.

gabe-l-hart:
Looks like this is a bad unit test from my previous branch. Will fix that independently and merge, then update this branch.

gabe-l-hart:
Test fix inbound: https://github.com/ibm-granite-community/granite-cli/pull/149

gabe-l-hart (approved):
Thanks for the contribution! This is a nice improvement to those silent errors. I've confirmed that the errors trigger as expected when the directory is unwritable and don't trigger when it is.

Once the bad test is fixed on `main`, I'll update this branch and merge.

PR #147, opened 2026-09-16
fix: create the bob workspace config directory before MCP registration

### LINKED ISSUE #144: bug: `launch bob` fails with an MCP enabled and no `.bob` directory
```sh
Fatal error: ENOENT: no such file or directory, open '/Users/ghart/Projects/github/ibm-granite-community/granite-cli-2/.bob/mcp.json'
Error: `/Users/ghart/.nvm/versions/node/v24.13.1/bin/bob mcp add-json vision-mcp {"headers":{},"type":"http","url":"http://127.0.0.1:57218/mcp"} -s workspace` exited with status exit status: 1
```

maskjelly:
Closes #144.

`bob mcp add-json … -s workspace` writes workspace-scoped MCP config to `<workspace>/.bob/mcp.json`, and the downstream `bob` binary expects that directory to exist — it does not create it itself. On a workspace that has never run bob before, registration dies with `ENOENT: … .bob/mcp.json` and the launch with it.

**The fix:** `BobLauncher::launch` now makes the directory (`ctx.working_dir/.bob`) immediately before the registration loop — only when there is at least one binding to register, and never on a dry run, which must not touch the filesystem. `create_dir_all` is a no-op when the directory is already there, so an existing `.bob/` is untouched, and the `remove` path is unchanged (it is best-effort by design and does not need the directory).

**One detail worth checking:** the guard tests the *final* binding list, after delegate sub-agents have pushed their `bob-sub-agents` binding onto it. A sub-agent-only launch therefore creates the directory too — it registers that server with the same `-s workspace` scope and would hit the same ENOENT otherwise. Testing the pre-delegate list would have left that half of the bug in place.

The directory name lives in a named constant with the why-comment, since it is a fact about bob's own layout rather than about this project's config.

**Tests:** three unit tests over the helper — a missing `.bob` is created, an existing one with an `mcp.json` in it is left byte-identical, and a dry run creates nothing. With the helper temporarily neutered the first fails (`assertion failed: … .bob is_dir()`); restored, 3/3 pass. `launch()` itself needs a real `bob` binary and has no fake-executable precedent in the repo's tests, so it stays uncovered end to end — the manual check on a machine with bob installed: `granite-cli launch bob` with an MCP binding on a fresh directory, confirming `.bob/mcp.json` appears and bob starts.

**Uncertainty, stated plainly:** bob resolves `-s workspace` against its process cwd, and this launcher spawns children without setting `current_dir` while `main.rs` builds `working_dir` from `current_dir()`, so `ctx.working_dir` is the workspace bob sees today. If bob ever resolves the workspace from a flag or its own config instead, this assumption should move with it.

**Gates** (CI's pinned 1.88.0): `cargo fmt --all -- --check`, `cargo clippy --all-targets --all-features -- -D warnings`, `cargo test --locked --all-targets --all-features` (879 passed) — all clean. Note for reviewers: the default toolchain is 1.98 here, whose clippy reports 5 pre-existing `collapsible_if` findings in other launchers; 1.88 is clean on both trees.


gabe-l-hart (approved):
Thanks for fixing this! Confirmed that it works as expected when `.bob` is missing and there are MCP servers to add. Also confirmed that `.bob` is _not_ created when there are no MCP servers to add.

Issue #59, opened 2026-08-14
refactor: reduce structural slop — tighten config/constructor contracts and establish clear layer boundaries


Tanker2020:
## Context

As the codebase has grown, several patterns have accumulated that make the code harder to read, test, and extend. None are bugs in isolation, but together they represent quality debt that compounds as new launchers and capabilities are added. This issue should be tackled **after** #58 lands, since #58 removes the deepest sources of these workarounds.

---

### 1. Typed outer fields vs. type-erased inner blob

Every `XxxConfig` struct carries both typed outer fields (`provider_id`, `provider_type`, `enabled_capabilities`) and a `config: serde_json::Value` inner blob. The outer fields are lost the moment only the blob is passed to a constructor. The fix is to pass the full typed struct (or a purpose-built view) rather than only the blob.

### 2. `unwrap_or_default()` swallows misconfiguration silently

Every `ConfigConstructable::new` implementation does:

```rust
let config: XxxConfig = serde_json::from_value(cfg.clone()).unwrap_or_default();
```

A malformed config silently falls back to defaults rather than producing a useful error. This should return `Result` and propagate the parse error to the caller.

### 3. `#[allow(unused)]` scattered through the factory macro

The `define_factory!` macro has a TODO comment acknowledging that several methods are annotated `#[allow(unused)]` as a placeholder. These should be removed as each method gains a real caller, or removed from the public surface if they genuinely serve no purpose.

### 4. Test doubles duplicated per-file

`FakeProvider`, `TestModelWithVariants`, and similar test doubles are defined inline in individual test modules. A `test_support` module (gated behind `#[cfg(test)]`) should centralise these so each new capability test doesn't copy-paste a full fake provider implementation.

### 5. Layer boundary violation: `Model::provider()` calling a global registry

The default `Model::provider()` impl in `models/base.rs` calls `PROVIDER_REGISTRY.construct(..., &Config::default())` — a model trait method reaching into a global static and constructing a fresh default config. This couples the `Model` layer upward into the provider registry and config layers. Provider resolution belongs in `ModelSource`, not on the trait itself.

### 6. `provider_config` sidecar injection

In several places, code builds an ad-hoc JSON blob like:

```rust
serde_json::json!({ "provider_config": provider_config })
```

[...]

PR #148, opened 2026-09-17
Running usage

gabe-l-hart:
## Description

This PR introduces the ability to track usage while a session is active via a new `sessions` tab in the interactive TUI.

## Changes

- [ARCHITECTURAL] Usage tracking is always on (no more opt-in)
  - The rationale for this is that usage tracking is one of the main value-adds for this project. The passthrough proxy is also active for multiple other uses already, so leaving it off by default doesn't save a whole lot.
- Usage tracking is now sync'ed to a file in `GRANITE_CLI_HOME/sessions` automatically during execution
- New TUI widget for viewing sessions (history and active) that auto-updates with usage as the session proceeds

## Testing

- Unit tests
- Manual testing by launching sessions an observing the state of the `sessions` yaml files and by observing the TUI panel

## Related Issue(s)

Closes #145

## AI Usage

This PR was built using a `granite-cli` config with Bob + all three named sub-agents (explore/granite4.2:3b, plan/Qwen3.6-35b, code/Qwen3.6-35b).

```sh
╔══════════════════════════════════════════════════════════╗
║           GIT AI USAGE ANALYSIS                          ║
╚══════════════════════════════════════════════════════════╝

📊 COMMITS BY AGENT

--- Aggregate ---
Commits                        |      Count
---------------------------------------------
IBM Bob + granite4.2:3b + Qwen3.6-35b |          4
IBM Bob                        |          1
IBM Bob + Qwen3.6-35b          |          1
---------------------------------------------
TOTAL                          |          6

📊 COMMITS BY USAGE TYPE

--- Aggregate ---
Commits                        |      Count
---------------------------------------------
none                           |          0
draft                          |          0
full                           |          6
---------------------------------------------
TOTAL                          |          6

📈 LINES OF CODE BY AGENT

--- Aggregate ---
Agent                     |    Commits |  Additions |  Deletions
------------------------------------------------------------
IBM Bob + granite4.2:3b + Qwen3.6-35b |          4 |       1147 |         58
IBM Bob                   |          1 |        115 |         36
IBM Bob + Qwen3.6-35b     |          1 |        197 |         12
------------------------------------------------------------
TOTAL                     |          6 |       1459 |        106

📈 LINES OF CODE BY USAGE TYPE

--- Aggregate ---
Usage Type           |    Commits |  Additions |  Deletions
-------------------------------------------------------
none                 |          0 |          0 |          0
draft                |          0 |          0 |          0
full                 |          6 |       1459 |        106
-------------------------------------------------------
[...]

PR #150, opened 2026-09-18
refactor: instance identity and reference resolution (part 1)

### LINKED ISSUE #51: feat: Instance Reference Semantics

### LINKED ISSUE #90: bug: launcher/provider/capability setup panics when a configured capability references a removed model

afrittoli:
## Description

First of three PRs implementing spec 0026. A capability, model or provider is
named in configuration by an id, and turning that id into an object was the
constructor's job: it received the whole application configuration, built every
configured instance of its kind to hand back one, and called `panic!` when a
name resolved to nothing.

This PR makes each source own the lookup for its own kind. A source records the
configuration it was built from and constructs an entry the first time that id
is asked for, keeping it so every later ask returns the same object. A model
reaches its provider by asking the provider source, and a capability reaches its
model by asking the model source, so nothing carries a copy of what it names.

Sub-tasks 4-10 follow in two further PRs: one construct signature, then one live
set of sources on the application context.

## Changes

- `efc9b0b` Add spec 0026 and a data model reference.
- `c675ed7` Sources hold, share and build on demand: instances are built on first ask, cached, and handed out as `Arc` so they outlive the source.
- `a1dd5e1` A model reaches its provider through the provider source: the copied `provider_config` blob, `Model::provider()` and `ProxiedModel` go; the session proxy moves down to `ProviderSource` and route registration to `src/proxy/routes.rs`.
- `6feeed0` A capability's model is resolved by the capability source: building a capability and wiring it to what it names become two steps, `ConfiguredModel::resolve`'s panic is deleted, and the capability's `ModelRequirement` is checked at resolve rather than only when setup picked the model.

## Testing

`scripts/lint.sh` and `cargo test` on the pinned toolchain (1.88.0). 881 tests
pass at the branch tip.

The suite is flaky under the default parallel test runner and clean with
`--test-threads=1`. Two launcher tests race on process-global state: several
tests in `opencode.rs` call `std::env::set_var`, and a `hermes.rs` dry-run test
asserts a shared `launcher_state_dir` is unchanged. Reproduced on the same
commits without this PR's changes, so it is not introduced here; tracked
separately.

## Related Issue(s)

- Closes #51 — one object per configured instance, held rather than rebuilt per call site.
- Closes #90 — a dangling reference is an error naming both ends instead of a panic, and denies service only to what names it.
- Part of #58 (the memoisation half; the single-argument constructor lands in the next PR) and #59 (items 5 and 6).

## AI Usage

Drafted with Claude Code (Opus 5) against the spec, reviewed and revised by
hand. Each commit records its own `AI-usage` trailer.


gabe-l-hart:
Reading through the description, this stands out to me as something for further discussion:

>The suite is flaky under the default parallel test runner and clean with
--test-threads=1. Two launcher tests race on process-global state: several
tests in opencode.rs call std::env::set_var, and a hermes.rs dry-run test
asserts a shared launcher_state_dir is unchanged. Reproduced on the same
commits without this PR's changes, so it is not introduced here; tracked
separately.

This kind of test race is one of the _main_ reason I always regret using singletons. I haven't reviewed the code yet, but the solution is usually to use context-singletons instead (ie singleton pattern tied to a non-singleton type like a `Config` object or a `Context` object that will be constructed once and shared in real usage, but can be constructed in isolation for tests).

docs/specs/0026-instance-identity-and-reference-resolution.md
```diff
@@ -0,0 +1,694 @@
...
+     │                                                                 │
+     │    providers   models   capabilities   launchers                │
+     │    built together, settings recorded, nothing constructed       │
+     │                                                                 │
+     │    ask "granite"   ──> constructed once ──> kept ──> a handle   │
+     │    ask "granite"   ─────────────────────────────────> the same  │
+     │                                                                 │
+     └── configuration write, or the session proxy starting ───────────┘
+              │
+              │   the set is discarded, with everything it built
```
gabe-l-hart, line 103: I'm a bit confused on this. It makes sense to me that when it writes to disk (ie something changes), the in-process config set is no longer valid and therefore discarded, but it seems like this would still allow the constructed instances based on the old config to persist as `Arc`s where they're being used. Couldn't this still strand the users with outdated configs?

docs/specs/0026-instance-identity-and-reference-resolution.md
```diff
@@ -0,0 +1,694 @@
...
+              ▼
+          the next ask builds a new set from the new configuration
+```
+
+The set is owned by `AppContext`, which hands out handles and discards it
```
gabe-l-hart, line 108: Ok, this _should_ mitigate the global-singleton-in-tests problem

docs/specs/0026-instance-identity-and-reference-resolution.md
```diff
@@ -0,0 +1,694 @@
...
+whenever configuration is written. A write is the only way to reach a mutable
+`Config`: reads go through `AppContext::config()`, writes through
+`AppContext::config_mut()`, and the second drops the set before it returns, so
+the borrow checker enforces the invalidation rule. The session proxy handle moves out of `Config` onto the context
+for the same reason, which makes starting the proxy one more event that ends a
+snapshot and rebuilds the set with providers that point at the local proxy.
```
gabe-l-hart, line 114: Interesting, this might have some consequences for the session-info I just added in #148

docs/specs/0026-instance-identity-and-reference-resolution.md
```diff
@@ -0,0 +1,694 @@
...
+
+A handle taken before a write keeps the instance it names alive, which is what
```
gabe-l-hart, line 116: Ok, this makes sense I think

docs/specs/0026-instance-identity-and-reference-resolution.md
```diff
@@ -0,0 +1,694 @@
...
+### Building an instance
+
+Construction is the same for every registry. A constructor reads its own
+settings and nothing else: no application configuration, no registry lookup,
+no network, no filesystem, no process spawn. What it can report is that its
+settings do not parse, or do not satisfy the validation its settings type
+declares, which is the report `unwrap_or_default` discards today.
+
+```rust
+fn new(instance_id: &str, cfg: &serde_json::Value) -> Result<Self, ConstructError>;
```
gabe-l-hart, line 130: There's an interesting hole that we might want to address here. The previous signature with `global_config` offered an escape hatch for the fact that `cfg` here is _only_ the `.config` generic json blob from the `<Type>Config` struct. Each of these structs has several additional fields that are expected to be common across _all_ implementations for the given trait family (eg. `ModelConfig.variant`). I think the right solution would be for `cfg` in `new` to be the top-level trait-type config (`ModelConfig`, `ProviderConfig`, etc). This would require some changes to `define_factory!` so that we can deduce the right type inside the macro. Currently, AFAIK, the only place that `<Type>Config` links with `trait <Type>` is in the `commands` where the `<Type>Config` is pulled from to-level `Config` to call the right registry's `construct`, so it's a loose coupling, not a tight type-based coupling.

docs/specs/0026-instance-identity-and-reference-resolution.md
afrittoli: Yes, indeed, using an `Arc` it is not enforced (at compile level), it is something consumers must take care of. In non-interactive mode or plain UI mode it is simple to ensure. 

If we are launching something, there's a chain of instances that is created, and they are needed for when the bind and actual launch happens. Right now there is nothing that runs in parallel to those, so the configuration cannot change; if there was something that could change the config (in the TUI, in future), the launch process would want to keep its original sources unchanged and not change mid-launch.

For the display use case in the TUI, it's important the opposite, i.e. to always have up to date config. 
That can be achieved by convention in the code. But even in the TUI, one might want to see the config which was used to start a running launcher.

Maybe the answer is to have two different ways into a source, one which returns an `Arc` which will survive config changes, and another one which returns a pointer that is invalidated the moment that config changes.

docs/specs/0026-instance-identity-and-reference-resolution.md
afrittoli: The problem with tests is not associated with instance, but with the `GRANITE_CLI_HOME` environment variable. I wonder if the few flakes I was were related to running manual testing in parallel to the automated ones 😅

docs/specs/0026-instance-identity-and-reference-resolution.md
afrittoli: Perhaps, yeah, worth checking

docs/specs/0026-instance-identity-and-reference-resolution.md
afrittoli: I wanted to explicitly separate the job of the constructor (map a cfg blob to an object) from that of looking up things, resolving references. As long as only the type specific `cfg` is the only thing in the constructor, we can enforce that.

If you think we may need in future additional type common fields, which are not related to resolving references, and which we don't want to put in the config blobs, with this proposal they would need to be set after construction. Another way of achieving it could be to have the `Config` module folding some common fields into every `cfg` object even if they are written in the file itself. 

Apart from the extra field issue, it may be worth tightening the loose coupling you mentioned, by making the factory type specific.

gabe-l-hart: This makes sense, and I think you're right that the common fields are almost entirely used for references. If we made that a strict requirement of common fields, that would be a good logical separation. I think the real issue I was thinking about when I did this was some of the really janky stuff going on with `ModelConfig` right now. There's some really broken logic in `build.rs` that highlights the fact that `type Config` is a `ConfigConstructable` is never actually validated against the json blob:

- All models get `type Config = crate::registry::NoConfig` ([here](https://github.com/ibm-granite-community/granite-cli/blob/main/build.rs#L78))
- In their constructors they look for `provider_config` in the json blob anyway ([here](https://github.com/ibm-granite-community/granite-cli/blob/main/build.rs#L82))
- `"provider_config"` gets shoved into the model's `cfg` blob in `ModelSource::from_config` ([here](https://github.com/ibm-granite-community/granite-cli/blob/main/src/models/mod.rs#L42))

Maybe if I actually read all the code here, this will already be solved? 😁

afrittoli: > * All models get `type Config = crate::registry::NoConfig`

What the model constructor could enforce there is that cfg is empty and not some random content.

> * In their constructors they look for provider_config in the json blob anyway
> `"provider_config"` gets shoved into the model's cfg blob in `ModelSource::from_config`

These should be solved by this branch.

docs/data-model.md
```diff
@@ -0,0 +1,75 @@
+# Data Model
+
+granite-cli reads two sets of data. The configuration holds the providers,
+models, capabilities and launchers a user has set up, as one YAML file per
+instance. The compiled-in data ships in the binary: the model catalogue
```
gabe-l-hart: NIT: `s/british/american/`

```suggestion
instance. The compiled-in data ships in the binary: the model catalog
```

src/capabilities/vision_mcp/mod.rs
```diff
@@ -96,15 +97,14 @@ impl ConfigConstructable for VisionMCPCapability {
     fn new(
         instance_id: &str,
         cfg: &serde_json::Value,
-        global_config: &crate::config::Config,
+        _global_config: &crate::config::Config,
```
gabe-l-hart, line 106: Hm, I thought we were getting rid of this entirely?

src/capabilities/vision_mcp/mod.rs
```diff
@@ -153,10 +162,11 @@ impl Capability for VisionMCPCapability {
         // support ImageUnderstanding, but the endpoint is looked up via
         // Chat, since that's the endpoint that actually serves vision
         // requests.
-        let (provider, endpoint, model_name) = self.configured_model.resolve_provider_endpoint(
+        let configured_model =
+            crate::capabilities::base::resolved_model(&self.configured_model, &self.instance_id)?;
+        let (provider, endpoint, model_name) = configured_model.resolve_provider_endpoint(
             model_id,
             ApiType::OpenAI,
-            ModelFunction::ImageUnderstanding,
```
gabe-l-hart, line 159: Hm, why is this removed? I do think we want `vision-mcp` to require `ImageUnderstanding`. All `Providers` that are capable of serving `Chat` w/ image inputs should also surface `ImageUnderstanding` as a supported `ModelFunction`.

src/capabilities/vision_mcp/mod.rs
```diff
@@ -493,19 +507,6 @@ mod tests {
         assert!(err.to_string().contains("no MCP transport in common"));
     }
 
-    #[tokio::test]
-    async fn bind_fails_when_model_lacks_image_understanding() {
```
gabe-l-hart, line 497: I think we'll want to keep this

src/models/base.rs
```diff
@@ -285,41 +272,34 @@ impl ConfiguredModel {
...
-        let provider = self
-            .model
-            .provider()
-            .map_err(|e| anyhow::anyhow!("model '{model_id}' has no usable provider: {e}"))?;
+        let provider = self.provider.clone();
         anyhow::ensure!(
             provider.supported_api_types().contains(&api_type),
             "provider for model '{model_id}' does not support {api_type}"
         );
-        anyhow::ensure!(
```
gabe-l-hart, line 317: Is this sort of requirement still enforced elsewhere? If not, we need to make sure we don't lose it. We shouldn't be sending image requests to a text model.

src/models/custom.rs
```diff
@@ -1,5 +1,5 @@
 // Third Party
-use alog::{MessageLevel, alog_channel, use_channel};
+use alog::use_channel;
```
gabe-l-hart: We can probably ditch `use_channel` if we don't need the others (ie there's no logging in this module).

src/models/custom.rs
```diff
@@ -83,21 +82,9 @@ impl ConfigConstructable for CustomModel {
         _global_config: &crate::config::Config,
     ) -> Self {
         let config: CustomModelConfig = serde_json::from_value(cfg.clone()).unwrap_or_default();
-        let provider_config = cfg.get("provider_config").and_then(|v| {
```
gabe-l-hart, line 86: Yes! This was such a hack

src/models/mod.rs
```diff
@@ -21,113 +21,209 @@ pub static MODEL_REGISTRY: LazyLock<base::ModelFactory> = LazyLock::new(|| {
...
-    /// connection details as a route on it (best-effort -- a registration
-    /// failure is logged and the model is returned untracked/unrouted rather
-    /// than failing construction over an accounting/routing feature) and
-    /// returns a model wrapped to point at the proxy instead of the real
-    /// upstream.
-    pub fn take(
-        &mut self,
+    /// Instance ids built so far, sorted. Lets a test tell what a call
+    /// built from what it merely could have built.
+    #[cfg(test)]
```
gabe-l-hart: I don't love test-only access (the old `friend` in c++), but I'm not sure of a better way to do this in rust.

src/models/mod.rs
```diff
@@ -21,113 +21,209 @@ pub static MODEL_REGISTRY: LazyLock<base::ModelFactory> = LazyLock::new(|| {
...
-                    "model '{model_id}' has no usable provider, skipping proxy route: {e}"
-                );
-            }
+        requirement: Option<&crate::capabilities::ModelRequirement>,
+    ) -> anyhow::Result<ConfiguredModel> {
+        let model = self.get(model_id)?;
+        if let Some(requirement) = requirement {
+            use crate::dependency::Requirement;
+            anyhow::ensure!(
+                requirement.admits_instance(&*model),
```
gabe-l-hart, line 148: Is this where `ModelFunction` support comes into play?

src/providers/mod.rs
```diff
@@ -21,48 +21,111 @@ pub static PROVIDER_REGISTRY: LazyLock<base::ProviderFactory> = LazyLock::new(||
...
+/// so every later ask for that id returns the same object.
 pub struct ProviderSource {
-    constructed: Vec<(String, Box<dyn Provider>)>,
+    /// The configuration this source was built from. Narrows to
+    /// `HashMap<String, ProviderConfig>` once `construct` stops taking the
+    /// whole configuration.
+    config: crate::config::Config,
+    /// When a session proxy is running, every provider handed out by `get`
+    /// points at it instead of the real upstream. Read from the
+    /// configuration for now; Sub-Task 5 has the launch pass it in directly.
```
gabe-l-hart: NIT: Slop comment referring to implementation implementation plan

gabe-l-hart (commented):
I made it through _most_ of the code, and I think it all makes sense. The general pattern is to replace all previous locations where instantiation was done by a call to `xyz_source.get`. This causes the fields in the structs to all become `Option`s that are not filled at construct time, but rather at dependency resolution time. I'm ok with this as the result of this change makes the rest of the code a lot tighter, though it does leave open the possibility of a misconfigured instance of a given type where the dependencies have not been resolved.

The two things I want to make sure are handled before approving are:

1. I _think_ all places that took `global_config` as part of `construct` now ignore it. If so, we should go the rest of the way and just remove this from the factory macros entirely.
2. I want to double-check that `ModelFunction` dependency gating is still supported. I _think_ I traced where that lives now, but want some confirmation.

src/models/base.rs
afrittoli: Yes, it was moved elsewhere, it is now separated from the binding.

The new `resolve_refs`, which happens before the binding, is what ensures that only a model that fits the requirements is used. if one of capabilities fails to resolve, the binding loop is not started and we avoid building things for nothing.

See this call chain: 

```rust
CapabilitySource::get
  └─ capability.resolve_refs(models)          // &mut self — the capability wires itself up
       └─ resolve_declared_model(models, &Self::metadata(), &self.config.model_id)
            └─ models.resolve(model_id, requirement)   // ModelLookup — the source turns a name into an object
                 ├─ self.get(model_id)                 // build or fetch from cache
                 ├─ requirement.admits_instance(...)   // check it satisfies what was declared
                 └─ ConfiguredModel::new(model, self.provider_for(model_id)?, variant)
```

src/capabilities/vision_mcp/mod.rs
afrittoli: Yes, in the next PR (sub-task 5).

src/capabilities/vision_mcp/mod.rs
afrittoli: A `self.configured_model` is only produced if that requirement was satisfied. If the requirement cannot be met, `bind` is not even reached anymore, we fail at resolution time.

See: https://github.com/ibm-granite-community/granite-cli/blob/1db42a393b2df84adfc0ca22a59fb713ce4be529/src/models/mod.rs#L155-L162

src/capabilities/vision_mcp/mod.rs
afrittoli: Uhm, this was removed because the capability check is now done *before* the bind. 
What this test does is inject an unvalidated model into bind, and with the new code it would fail.

While the code from he PR always executes resolution before the binding, there's nothing in the API that enforces it, so it's a gap of this approach. 

I could build something like:

```rust
trait Capability {
    fn resolve_refs(self: Box<Self>, models: &dyn ModelLookup)
        -> anyhow::Result<Box<dyn ResolvedCapability>>;
}

trait ResolvedCapability {
    async fn bind(&self, request: BindingRequest) -> anyhow::Result<Binding>;
    async fn on_setup(&self) -> anyhow::Result<()>;
    // ...
}
```

Which would embed in the types how the API must be used.

src/models/mod.rs
afrittoli: Uhm, there probably is, let me address is in a follow-up commit.

src/models/mod.rs
afrittoli: Yes, exactly

gabe-l-hart:
@afrittoli Looks like we need `./scripts/fmt.sh` to get CI to pass

src/capabilities/vision_mcp/mod.rs
gabe-l-hart: Right, sorry, this was as I was thinking the logic for `ModelFunction` filtering had been lost.

afrittoli:
> I made it through _most_ of the code, and I think it all makes sense. The general pattern is to replace all previous locations where instantiation was done by a call to `xyz_source.get`. This causes the fields in the structs to all become `Option`s that are not filled at construct time, but rather at dependency resolution time. I'm ok with this as the result of this change makes the rest of the code a lot tighter, though it does leave open the possibility of a misconfigured instance of a given type where the dependencies have not been resolved.
> 
> The two things I want to make sure are handled before approving are:
> 
> 1. I _think_ all places that took `global_config` as part of `construct` now ignore it. If so, we should go the rest of the way and just remove this from the factory macros entirely.

Yes, that's coming in the next PR.

> 2. I want to double-check that `ModelFunction` dependency gating is still supported. I _think_ I traced where that lives now, but want some confirmation.

Yes, absolutely, it's now happening before `bind` via `resolve_ref`.

docs/specs/0026-instance-identity-and-reference-resolution.md
gabe-l-hart: Looks like tracking is still working fine

gabe-l-hart (approved):
Ship it!

